PR #112, opened 2026-04-17
feat: add in-memory pricing catalog for LLM cost reporting

alenkacz:
## Summary

Adds an in-memory pricing catalog for LLM cost reporting in the AI Gateway. Pricing is embedded directly in each provider's `ModelDefinition` — when you add a model, you define its pricing alongside capabilities and constraints. This makes it impossible for models and pricing to drift apart.

### What's in the catalog

| Provider | Models | Notes |
|----------|--------|-------|
| **OpenAI** | 21 | GPT-5.x family, GPT-4.x, o-series |
| **Anthropic** | 7 | Claude Sonnet/Opus/Haiku 4.1–4.7, with cache write pricing (5m/1h TTL) |
| **Google** | 7 | Gemini 2.0–3.1, tiered pricing for Pro models (≤200k vs >200k context) |
| **Bedrock** | 6 | Claude on Bedrock (uniform pricing across regions), with cache write pricing |

### Key design decisions

- **Microcents (int64)** — All prices stored as microcents per million tokens (1 cent = 1,000,000 microcents). Avoids float precision loss across cost aggregation pipelines. Sub-cent prices like \$0.005/M are represented exactly as \`500_000\`.
- **Pricing embedded in ModelDefinition** — \`ModelPricing()\` per provider derives from \`supportedModels\`, so it's impossible for pricing and models to drift apart. \`TestAllModelsHavePricing\` in each provider fails CI if any model has zero pricing.
- **Tiered pricing** — Google Gemini Pro models charge differently based on context length (≤200k vs >200k). \`Info.Tiers\` + \`CalculateCost()\` select the correct tier based on \`inputTokens + cachedTokens\`.
- **Anthropic cache write pricing** — \`AnthropicPricing\` struct with \`CacheWrite5mPerMillion\` (1.25× input, default) and \`CacheWrite1hPerMillion\` (2× input, extended). \`CacheWriteRate(ttl)\` accessor returns the correct rate for \`\"5m\"\` or \`\"1h\"\`. Returns 0 for non-Anthropic models (OpenAI cache writes are free, Google charges per-hour storage).
- **Catalog version hash** — \`Catalog.Version()\` returns a deterministic SHA256 content hash (16 hex chars). Used to tag spending events so you can determine which catalog version priced a given event.
- **\`EffectiveFrom\`** — Supports historical rate lookups. When multiple rates exist for a model, the correct rate is selected by comparing \`EffectiveFrom\` against the event timestamp.

### Pricing accuracy gaps (known, documented)

- **Cache TTL extraction** — The hook defaults to 5m TTL. Extracting actual TTL from request body is deferred (P2).
- **Region-based pricing** — Bedrock Anthropic models have uniform pricing across regions. Non-Anthropic Bedrock models (future) may have regional differences.
- **Priority/speed variants** — Anthropic \`speed: \"fast\"\` currently has identical pricing. No impact today.
- **Bedrock model coverage** — Only Anthropic Claude models. Llama, Mistral, etc. are not yet covered.
- **Staleness detection** — Planned: integration tests that query provider model-listing APIs weekly and open issues when new models appear.

## Test plan

- [x] \`pricing/\` package: catalog lookup, cost calculation, tiered pricing, cache write pricing (5m/1h), nil Anthropic handling, version hash determinism, multi-provider merging
- [x] \`TestAllModelsHavePricing\` per provider — every model must have non-zero input/output pricing
- [x] Anthropic/Bedrock tests enforce \`AnthropicPricing\` sub-struct with both cache write rates
- [x] \`TestModelPricingMatchesModels\` per provider — 1:1 coverage between \`supportedModels\` and \`ModelPricing()\`
- [x] Verified integration from cloudv2 aigw spending hook via replace directive


alenkacz:
@claude review

pricing/pricing.go
```diff
@@ -0,0 +1,143 @@
...
+package pricing
+
+// Info holds pricing for a model. Embed this in provider ModelDefinition
+// structs so pricing is defined alongside capabilities.
+// All prices are in microcents per million tokens.
+//
+// For tiered models, set only Tiers — the flat fields (InputPerMillion etc.)
+// are automatically populated from the first (lowest) tier when building
+// a Catalog. For flat-rate models, set the flat fields and leave Tiers nil.
+type Info struct {
```
weeco, line 53: There are a couple more provider specific costs. How can we model them? For example:
- Regions
- Variants such as fast/priority or batch
- Explicit Cache writes for different durations

alenkacz:
@claude review

pricing/pricing.go
alenkacz: we agreed on omitting regions. I added fast, we don't support batch right now (we can add later). Added cache writes (anthropic specific)

birdayz (approved):

birdayz:
+1 / merged to unblock - if there's any further issues, let's solve it afterwards

PR #116, opened 2026-04-21
fix: surface truncation instead of corrupting state when tool_use is cut off

birdayz:
## What

When a streaming tool_use block is cut off mid-args (typically `stop_reason=max_tokens` during parallel tool use), surface the truncation instead of silently corrupting session state.

## Why

Two bugs, one unified wedge:

1. Providers accumulating tool args delta-by-delta (anthropic's `input_json_delta`, bedrock's `toolUse.input`, openaicompat's `tool_calls[].function.arguments`) were emitting the block regardless of whether the stream had actually finished accumulating. Partial JSON like `{"query":` got handed back as `ToolRequest.Arguments`, persisted to session state, and poisoned every subsequent replay with `unexpected end of JSON input`. Confirmed in production.

2. Every provider's `FinishReason` logic said "tool calls take precedence": if any `tool_use` block made it to the response, `FinishReason` got clamped to `ToolCalls` even when the actual stop reason was `max_tokens`. Agent loops never saw the truncation signal, so they kept executing the N-1 completed calls and the model had to guess why the Nth call vanished.

### Drop partial tool_use at stream finalization

Providers susceptible to the accumulation bug — anthropic, bedrock, openaicompat — got a small `finalizeToolArgs` / `finalizeToolRequest` helper. Empty accumulation coerces to `{}` (providers' wire form for no-arg tool calls). Invalid JSON signals a cut-off block and the caller drops it from both the mid-stream `ContentPartEvent` path and the stream-end reconstruction.

Audit of the other providers:

- **openai Responses**: only emits tool calls on `response.output_item.done`, which doesn't fire for partial items. Not susceptible.
- **google/gemini**: `FunctionCall.Args` arrives as a pre-parsed `map[string]any` from the Gemini SDK. Not susceptible.

### Stop lying about FinishReason

Every provider (anthropic, bedrock, google, openai, openaicompat) had the same anti-pattern: `if hasToolCalls { FinishReasonToolCalls } else { mapStopReason(...) }`. Now the mapped stop reason runs first, and the `ToolCalls` upgrade only kicks in when the mapped reason was a plain `Stop`. Truncation (`Length`), content filter, and other non-stop signals propagate through untouched.

The agent loop already handles `FinishReasonLength` correctly — `agent/llmagent/llmagent.go:295` emits a status event and terminates the loop. That code path was just unreachable because the override clobbered the signal.

### What happens after the fix

Parallel tool use hits `max_tokens` on the Nth call:

1. Provider drops the partial Nth block at stream finalization.
2. `FinishReason` comes through as `Length` (because the mapped stop reason is `Length` and we no longer force `ToolCalls`).
3. `mapLLMFinishReason` in the agent loop returns `FinishReasonLength` (terminal).
4. Agent emits a "turn N completed - length limit" status event and stops.
5. Tools are **not** executed. The completed N-1 tool_use blocks live in `Response.Message.Content` for the caller to inspect, but the loop terminates cleanly.
6. Caller decides the policy: retry with higher `max_tokens`, surface to user, whatever.

This is honest. Executing N-1 out of N intended calls and pretending everything is fine is worse than stopping and letting the caller retry.

## References

Martin pointed out bedrock needed the same stream-side fix (earlier PR comment). Johannes pointed out we were still swallowing the signal by clamping `FinishReason` to `ToolCalls`. Both addressed in this single commit.


weeco:
@claude Review this PR and check why the bedrock provider doesn't have a fix for the streaming handling - was that missed?

providers/bedrock/request_mapper.go
```diff
@@ -380,3 +374,22 @@ func (rm *RequestMapper) mapToolChoice(choice *llm.ToolChoice) (types.ToolChoice
...
 	}
 }
+
+// decodeToolArgs converts a persisted ToolRequest.Arguments payload into the
+// JSON object shape Bedrock's tool_use input requires. See the matching
+// helper in providers/anthropic for the full rationale; the short version is
+// that streaming turns cut off mid-tool-call (e.g. stop_reason=max_tokens)
+// persist truncated JSON, and without this fallback replay is permanently
+// wedged on "unexpected end of JSON input".
+func decodeToolArgs(arguments json.RawMessage) map[string]any {
```
weeco: I guess we can move them to a shared helper somewhere since this func is necessary for all providers?

birdayz: Good call. Moved to `llm.ToolRequest.ArgumentsAsObject()` in 0acb2c5c — anthropic, bedrock, and google all call through the shared method now. openaicompat's `finalizeToolRequest` stays local; different semantics (it gates emission rather than decoding for a structured API) and only one caller.

birdayz:
Fixed bedrock in e78a4b3c. Same approach as the anthropic and openaicompat fixes: replaced the coerce-invalid-to-{} behavior of `argsJSON()` with a `finalizeToolArgs() (bytes, ok)` helper that returns `(nil, false)` on truncation, and both emission sites drop the block when finalize rejects. Partial tool_use no longer leaks into the final response.

weeco (approved):

birdayz:
@claude plz review

PR #122, opened 2026-04-28
feat(bedrock): split models by inference profile

alenkacz:
## Summary

Split each Claude model on Bedrock into one catalog entry per AWS inference profile, so `Models()` and downstream cost reporting attribute per-profile (matching how LiteLLM and TrueFoundry expose Bedrock) and so the catalog never advertises a model ID that AWS won't accept on-demand.

## Why

The previous single-entry catalog had two problems:

1. **`Models()` exposed un-invokable IDs.** It returned bare names like `anthropic.claude-opus-4-7`, but newer 4.6+ models are inference-profile-only on AWS — calling them directly returns *"Invocation … with on-demand throughput isn't supported. Retry your request with the ID or ARN of an inference profile that contains this model."*
2. **Spend attribution couldn't distinguish profiles.** Every request routed through the SDK was billed by AWS as a specific inference profile (`us.` / `global.` / etc.), but the catalog only had a single logical entry per model — the gateway couldn't tell `us.opus-4-7` apart from `global.opus-4-7` in cost reports.

## What changed

- **`providers/bedrock/models.go`** — restructured catalog. Each logical Claude model is described once via a `claudeModel` struct (capabilities, constraints, pricing, supported geo profiles, whether it's inference-profile-only). `expand()` produces one `ModelDefinition` per profile (`global.` + each geo). Per-variant geo coverage cross-checked against AWS's per-model card pages, not LiteLLM (which has gaps and stale entries — we found `us.` missing for Opus 4.5, `jp.` missing for Sonnet 4.5, and a phantom `apac.` entry for Haiku 4.5 that AWS doesn't actually publish).
- **Bare entries are dropped for inference-profile-only models** (Sonnet 4.6, Opus 4.6, Opus 4.7). `Models()` won't list `anthropic.claude-opus-4-7` because it isn't on-demand invocable; the bare Go constant remains as a building block for the prefixed variants and as the ergonomic input form `NewModel("anthropic.claude-opus-4-7")` accepts (auto-prefix → matching geo entry).
- **`providers/bedrock/provider.go`** — `NewModel()` looks up the prefixed `apiModelID`, not the raw input, so the resolved `ModelDefinition` matches the actual inference profile AWS will route to.
- **`lookupModel`** is exact-match only — no prefix-stripping fallback (each variant is its own entry).
- **`response_mapper.go`** (via test) — `InvokedModelID` now reflects the actual inference profile AWS reported, not the bare logical name. Downstream cost attribution depends on routing identity.
- **Conformance fixture** — filters `Models()` to entries with the test region's geo prefix. Geo/`global.` variants stay in the SDK catalog (consumers see them) but are not exercised live, since one CI region can't reach all geographies and the SCP denies the foundation-model ARNs that `global.` resolves to. Per-variant catalog correctness is unit-tested.
- **`InferenceProfileRegion`** is now exported so the conformance fixture can compute the test region's geo prefix without duplicating logic.

## Known limitation: pricing is still flat per logical model

This PR keeps a single flat rate per logical Claude model, applied uniformly to every inference profile variant. AWS actually charges a ~10% premium on the geo profiles (`us.` / `eu.` / `au.` / `jp.`) over the headline rate, so cost reports under-report by ~10% for any geo invocation. Since `NewModel()` auto-prefixes a geo when given a bare name, that's the common path.

I had this fixed in an earlier revision of this PR (per-variant rates with the +10% premium baked in) but reverted it: the proper fix also has to handle non-Anthropic Bedrock models — Nova, Gemma, DeepSeek — where regional pricing can differ by tens of percent in non-uniform ways, which wants the existing `pricing.Info` `Selector{Region: …}` machinery rather than a per-prefix rate field. Tracked in [AI-957](https://redpandadata.atlassian.net/browse/AI-957).

## Pattern parity

Mirrors how LiteLLM and TrueFoundry expose Bedrock — every variant is its own catalog row, billed and reported separately.

## Backward compatibility

- Public Go API surface unchanged: `NewProvider`, `NewModel`, `WithRegion`, `ModelPricing`, `With*` options. `InferenceProfileRegion` is newly exported.
- Existing callers that pass un-prefixed names (`NewModel(ModelClaudeSonnet46)`) still work — auto-prefix is unchanged.
- `Models()` now returns ~30 entries instead of 6.
- `apac.` is no longer in the catalog for any Claude model — it never existed on AWS for these models, was an incorrect carry-over from LiteLLM.

## Test plan

- [x] `go test ./providers/bedrock/... -short` passes
- [x] `go vet ./...` clean
- [x] golangci-lint (godoclint, wsl_v5) clean
- [ ] CI integration tests (`TestBedrockConformance_Integration`) — fixture now filters to the test region's geo prefix so we don't try to invoke profiles the CI can't reach
- [ ] Reviewer sanity-check that the variant set per model matches AWS's per-model card pages (catalog was rebuilt against those, not LiteLLM)

[AI-957]: https://redpandadata.atlassian.net/browse/AI-957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/providers/bedrock/bedrock_conformance_test.go
+++ b/providers/bedrock/bedrock_conformance_test.go
@@ -17,6 +17,7 @@ package bedrock_test
 import (
 	"context"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/redpanda-data/ai-sdk-go/internal/testsuite"
@@ -83,7 +84,34 @@ func (f *BedrockFixture) NewReasoningModel(t *testing.T) llm.Model {
 }
 
 func (f *BedrockFixture) Models() []llm.ModelDiscoveryInfo {
-	return f.provider.Models()
+	all := f.provider.Models()
+
+	// The catalog now exposes one entry per inference profile (base, global.,
+	// us./eu./au./apac.). Conformance runs in a single AWS region with limited
+	// IAM, so only the bare entries are universally reachable: NewModel
+	// auto-prefixes them to the test region's geo profile, which is the only
+	// routing the CI's SCP allows. Geo profiles for other geographies fail with
+	// ValidationException (cross-geography from a single entry-point region),
+	// and the global. profile resolves to foundation-model ARNs that the CI's
+	// SCP explicitly denies. Per-variant catalog correctness is covered by
+	// unit tests in provider_test.go.
+	filtered := make([]llm.ModelDiscoveryInfo, 0, len(all))
+
+	for _, m := range all {
+		if isBareModelID(m.Name) {
+			filtered = append(filtered, m)
+		}
+	}
+
+	return filtered
+}
+
+// isBareModelID reports whether name is a foundation-model ID without an
+// inference-profile prefix (e.g. "anthropic.claude-opus-4-7" yes;
+// "us.anthropic.claude-opus-4-7" no).
+func isBareModelID(name string) bool {
+	return strings.HasPrefix(name, "anthropic.") &&
+		!strings.Contains(strings.TrimPrefix(name, "anthropic."), ".")
 }
 
 func (f *BedrockFixture) NewModel(modelName string) (llm.Model, error) {
--- a/providers/bedrock/bedrock_conformance_test.go
+++ b/providers/bedrock/bedrock_conformance_test.go
@@ -31,6 +31,7 @@ import (
 // BedrockFixture implements the conformance.Fixture interface for the Bedrock provider.
 type BedrockFixture struct {
 	provider *bedrock.Provider
+	region   string
 }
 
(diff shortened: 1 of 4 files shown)


simon0191 (approved):

malinskibeniamin (approved):
I think inference region is an implementation detail, right? Would it be something we still want to expose in the UI?

alenkacz:
@malinskibeniamin it's not really, because you cannot call the model without the inference profile in name if you have on-demand pricing.

PR #126, opened 2026-05-02
fix(bedrock): reject tool with empty description before calling Converse

simon0191:
## Summary

- Bedrock's Converse API rejects any tool whose `description` is the empty string with an opaque `ValidationException` (`Member must have length greater than or equal to 1`), failing the entire call. The mapper used to pass empty descriptions through unchanged, so callers got a server-side error that pointed at tool indices (`toolConfig.tools.12.member.toolSpec.description`) rather than tool names — hard to map back to source for an agent that mixes MCP-derived and subagent-as-tool entries.
- `mapToolConfig` now validates upfront and returns one error listing every offender by name, e.g. `bedrock: every tool must have a non-empty description; missing on: subagent_a, subagent_b`. Errors bubble up wrapped in `llm.ErrRequestMapping`.
- Scoped to the Bedrock provider deliberately. OpenAI and Anthropic accept empty tool descriptions, so a check at `agent.New()` would crashloop agents that work fine on those providers.

## Context

Hit in production by an AI agent using subagents whose `Subagent.description` was empty (proto schema marks it `OPTIONAL` with no `min_len`). On the first Bedrock call the whole request failed; the cloudv2 side will land a separate proto-level fix to make `description` non-empty for new data, but this client-side preflight gives every existing caller a useful error today.

## Test plan

- [x] `go test ./providers/bedrock/...` — full provider suite passes
- [x] `golangci-lint run ./providers/bedrock/...` — clean
- [x] New `TestRequestMapper_EmptyToolDescription` asserts: error wraps `llm.ErrRequestMapping`, both empty-description tools are named in the message, valid tools are not flagged

### DIFF
--- a/providers/bedrock/request_mapper.go
+++ b/providers/bedrock/request_mapper.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/bedrockruntime"
@@ -314,6 +315,22 @@ func (rm *RequestMapper) mapToolResultBlock(resp *llm.ToolResponse) types.ToolRe
 
 // mapToolConfig converts tool definitions and choice to Bedrock ToolConfiguration.
 func (rm *RequestMapper) mapToolConfig(tools []llm.ToolDefinition, choice *llm.ToolChoice) (*types.ToolConfiguration, error) {
+	// Bedrock Converse rejects any tool whose description is the empty
+	// string with an opaque ValidationException ("Member must have length
+	// greater than or equal to 1"). Validate upfront and surface every
+	// offender at once so callers can fix all of them in one pass.
+	var missingDesc []string
+
+	for _, tool := range tools {
+		if tool.Description == "" {
+			missingDesc = append(missingDesc, tool.Name)
+		}
+	}
+
+	if len(missingDesc) > 0 {
+		return nil, fmt.Errorf("bedrock: every tool must have a non-empty description; missing on: %s", strings.Join(missingDesc, ", "))
+	}
+
 	apiTools := make([]types.Tool, 0, len(tools))
 
 	for _, tool := range tools {
--- a/providers/bedrock/provider_test.go
+++ b/providers/bedrock/provider_test.go
@@ -470,6 +470,42 @@ func TestRequestMapper_ToolDefinitions(t *testing.T) {
 	assert.True(t, ok)
 }
 
+func TestRequestMapper_EmptyToolDescription(t *testing.T) {
+	t.Parallel()
+
+	cfg := &Config{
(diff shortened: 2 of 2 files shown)


weeco:
@claude im wondering whether we should do the check at tool registration time. Would it make sense to require this for all tools across providers? Breaking changes / behaviors are still acceptable

birdayz:
hah i had exactly same thought @weeco . agree, it should be done at registration time.

birdayz:
i digged a bit deeper, and maybe it's the wrong call to reject empty desc always.. 

```
  ┌─────────────────────┬─────────────────────────────────────────────────────────────────────────────┐
  │         SDK         │                           Validates description?                            │
  ├─────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
  │ adk-go              │ No — Tool.Description() string interface, value is passed through unchecked │
  ├─────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
  │ adk-python          │ No — BaseTool accepts description: str with no runtime check                │
  ├─────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
  │ openai-agents-js    │ No — typed as string in StrictToolOptions, no runtime check                 │
  ├─────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
  │ ai-sdk-go (this PR) │ Bedrock-only at mapping time                                                │
  └─────────────────────┴─────────────────────────────────────────────────────────────────────────────┘
  
    Nobody else validates this — anywhere. Not at registration, not at model-call, not provider-specifically. Type systems make it string,
   but empty string passes.
  ```
  
IMHO would not be right call to reject at registration time, the runtime fix is cleaner. LGTM

**Review by birdayz — APPROVED** [review]

File: providers/bedrock/request_mapper.go [hunk 1]
```diff
@@ -314,6 +315,22 @@ func (rm *RequestMapper) mapToolResultBlock(resp *llm.ToolResponse) types.ToolRe
...
+
+	for _, tool := range tools {
+		if tool.Description == "" {
+			missingDesc = append(missingDesc, tool.Name)
+		}
+	}
+
+	if len(missingDesc) > 0 {
+		return nil, fmt.Errorf("bedrock: every tool must have a non-empty description; missing on: %s", strings.Join(missingDesc, ", "))
+	}
```
weeco, line 332: @simon0191 I was bothered by this solution and did some research. I noticed:

1. Bedrock / AWS itself claims the tool description is NOT required, but when provided it can't be empty, see: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolSpecification.html
2. This leads me to believe that this is the problem - we should omit it instead of setting an empty `aws.String()`

Have you tried this and see if that works?

weeco (changes requested):

providers/bedrock/request_mapper.go
birdayz: good catch!

PR #131, opened 2026-05-09
Add Vercel AI SDK UI Message Stream adapter

birdayz:
## Summary

Adds `adapter/vercelaisdk/uimessagestream/` — an HTTP handler that implements the Vercel AI SDK UI Message Stream protocol (v1). Any `llm.Model` can now serve responses compatible with `useChat` from `@ai-sdk/react` v6.

- **Text streaming**: `start → start-step → text-start → text-delta* → text-end → finish-step → finish → [DONE]`
- **Tool calls**: `tool-input-start → tool-input-available → tool-output-available` with agentic loop (`StreamModelWithTools`)
- **Reasoning traces**: `reasoning-start → reasoning-delta* → reasoning-end` with stateful tracking
- **Error handling**: sanitized error text to clients, proper `finish-step + finish(error)` sequence
- **Security**: 1MB request body limit, messageId generation via crypto/rand

### Wire format

1:1 port of the Vercel AI SDK's `toUIMessageStream()`. Every chunk passes Zod `z.strictObject` validation — verified field by field against the TypeScript schemas.

### API

```go
mux.Handle("/api/chat", uimessagestream.Handler(model,
    uimessagestream.WithSystem("You are helpful."),
    uimessagestream.WithTools(tools, executor),
))
```

### Testing

- 30 Go unit tests (87.8% coverage)
- 6 JS conformance tests using real `ai@6.0.177` `DefaultChatTransport` against a Go test server
- Reviewed by 3 expert personas (Go, Vercel protocol, DevOps) across 2 rounds

## Test plan

- [x] `go test ./adapter/vercelaisdk/uimessagestream/ -v` — 30 tests pass
- [x] `cd adapter/vercelaisdk/uimessagestream/conformance && ./run.sh` — 6 JS conformance tests pass
- [x] Verified end-to-end with Alfred chat app (text, tool calls, reasoning, model selection)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #132, opened 2026-05-11
vercelaisdk: target ai@7.0.6 and fix conformance bugs

birdayz:
## What

Bring the Vercel AI SDK UI message stream adapter to full conformance with `ai@7.0.6` and fix the bugs that kept its conformance suite from actually passing. Adds an error mapper, reconstructs inbound tool-call history, and expands the test + conformance coverage.

## Why

The adapter claimed conformance but the suite did not pass: the error path emitted a hardcoded "An error occurred" while the test -- and the reference contract -- expected the real error to be surfaceable. Multi-turn tool conversations were silently dropped because only text was forwarded to the model. Smaller deviations from the reference (finish reason on turn-limit exhaustion, span/error ordering, no abort signal on cancellation, HTML-escaped wire bytes) would surface with any non-trivial model behavior. The v6->v7 protocol change is purely additive -- chunk shapes, SSE framing and headers are unchanged -- so the bump is safe.

## Implementation details

**Error handling.** Add `WithOnError(func(error) string)`, mirroring the reference `onError` option. The default returns the sanitized "An error occurred." (with the trailing period the reference uses). Every error chunk -- terminal, iterator, recoverable `ErrorEvent`, and `tool-output-error` -- routes the real error through the mapper instead of a hardcoded string. The model still receives the real tool error; only the client-facing text is mapped.

**Inbound history.** Reconstruct assistant tool calls and their results from the UI message parts (`tool-<name>` / `dynamic-tool`), split on step boundaries, mirroring `convert-to-model-messages.ts`. Coalesce consecutive same-role messages so roles stay strictly alternating -- the reference leaves this to the provider, and the Go providers do not coalesce. Tool results carry their payload in the sealed-interface `ToolResponsePart` (`IsError` + `Result`). Previously only text was forwarded and multi-turn tool conversations were dropped.

**Other fixes.**
- `maxTurns` exhaustion emits the last turn's real finish reason (`tool-calls`), not a synthetic `other`.
- `closeSpans()` runs before the error chunk on every terminal path, so an error is never wedged inside an open text run.
- Emit an `abort` chunk + `[DONE]` on context cancellation, matching `stream-text.ts`.
- Disable HTML escaping in the SSE encoder so `<`, `>` and `&` in deltas match `JSON.stringify` byte-for-byte.

**Conformance and tests.** Bump the suite to `ai@7.0.6` and add endpoints + tests for reasoning, tool calling, tool errors, multi-step tool loops and mid-stream errors; it now passes 11/11 against the real client, which strict-schema-validates every chunk. Add a `task test:conformance` target and Go-level schema + parse-state-machine invariant checkers.

## References

Original code review: https://github.com/redpanda-data/ai-sdk-go/pull/131#issuecomment-4419245214

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ go.work.sum
 .idea
 .build
 .bin
+
+# Vercel AI SDK reference clone (for adapter cross-checking)
+/vercel-ai-reference*/
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -73,6 +73,12 @@ tasks:
     cmds:
       - 'GOROOT={{.GO_ROOT}} GOTOOLCHAIN=local {{.GO_BIN}} test -json -cover -parallel 10 -timeout 30m -run Integration ./... | {{.BIN_DIR}}/tparse -follow -all'
 
+  test:conformance:
+    desc: Run the Vercel AI SDK UI message stream conformance suite (requires node + bun)
+    dir: adapter/vercelaisdk/uimessagestream/conformance
+    cmds:
+      - bash run.sh
+
   security:
     desc: Run security vulnerability checks (govulncheck + osv-scanner)
     deps:
--- a/adapter/vercelaisdk/uimessagestream/conformance/bun.lock
+++ b/adapter/vercelaisdk/uimessagestream/conformance/bun.lock
@@ -5,24 +5,24 @@
     "": {
       "name": "aisdk-conformance",
       "dependencies": {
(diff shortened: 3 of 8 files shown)


birdayz:
@claude review

birdayz:
@claude review

birdayz:
@claude review

birdayz:
@claude review

birdayz:
@claude review

malinskibeniamin (approved):

PR #133, opened 2026-05-11
mcp: enforce 64-char tool name limit after server ID prefixing

birdayz:
## What

Enforce the 64-character tool name limit in the MCP client's `namespaceTool()` by mangling names that exceed the limit after the `serverID__` prefix is applied.

## Why

`namespaceTool()` concatenates `serverID__toolName` without checking length. LLM providers (at minimum Bedrock) reject tool names exceeding 64 characters. protoc-gen-go-mcp already mangles proto-generated names to 64 chars, but the server ID prefix pushes them right back over:

- `servicenow-catalog__delete_api_sn_sc_servicecatalog_cart_by_sys_id_empty` (73 chars)
- `servicenow-catalog__post_api_sn_sc_servicecatalog_items_by_sys_id_versioning_checkout` (85 chars)

## Implementation details

Reuses the same mangling algorithm from protoc-gen-go-mcp: SHA-256 the full name, take 10 base-36 chars as a deterministic hash prefix, keep as much of the tail (the most human-readable part) as fits within 64 chars.

Names within the limit pass through unchanged -- zero behavioral change for short names.

Since mangled names cannot be reversed via `strings.TrimPrefix`, `toolWrapper` now stores the original server tool name. `ExecuteTool()` looks it up from the wrapper, falling back to `TrimPrefix` for direct API callers.

## References

AI-1106


birdayz:
@claude plx review

alenkacz (approved):

birdayz:
Addressed the review in cb190d64:

- **#1** Entropy comment corrected: ~49 bits after leading-digit bias, 50% birthday bound ~24M.
- **#2** Upgraded from the suggested warn-log to a hard error: `prepareTools` now fails the whole sync on a namespaced-name collision, keeping the previous tool set intact instead of silently dropping a tool. Detection happens at sync time, the earliest point the names exist. Not unit-testable without a seam (requires an actual SHA-256 prefix collision; the fetched map already dedupes raw names, and unmangled names always start with `serverID__` while mangled ones start with a base36 hash digit) — it's a tripwire.
- **#3** ListTools doc updated as suggested.
- **#4** Kept the fallback (it serves hand-constructed plain names), added the comment explaining why and its limits.
- **#5/#7** ASCII assumption and minimum-useful-maxLen documented in the godoc.
- **#6** Added `TestIntegrationExecuteToolMangledName`: lists a mangled name, dispatches it, asserts the server receives the original tool name.

malinskibeniamin (approved):

birdayz:
@claude plz re-review

PR #137, opened 2026-05-19
fix(runner): avoid range-over-func panic on per-message save error [AI-1157]

MikolajGagatekRP:
## What

In `runner/runner.go`, the per-`MessageEvent` inner session save (line 218) called `yield(nil, err)` without checking its bool return. If the consumer broke on that error, the deferred final save then re-entered `yield` on a canceled iterator and Go panicked with `range function continued iteration after function for loop body returned false`.

Surgical fix: capture `yield`'s return at that site; set `consumerStopped = true` on `false` so the existing defer-time guard (lines 192-198) fires.

## Why

[AI-1157](https://redpandadata.atlassian.net/browse/AI-1157) — observed deterministically in `marketing-agent` (Cloud Run worker pool on `rp-internal-agents`) when an LLM returns a malformed `json.RawMessage` in a tool-args block. `MarshalJSON` on the next session save fails, the consumer breaks on the yielded `ErrSessionSave`, and the deferred save's yield panics — crashing the worker before it can emit the Slack reply.

Distinct from AI-587 (consumer breaks on `InvocationEndEvent`), which was already handled because that yield site at line 224 already captured the bool.

## How

- `runner/runner.go`: capture the bool return at line 218 and set `consumerStopped` on `false`.
- `runner/runner_test.go`: new `TestRun_SessionSaveError_ConsumerBreaksOnMessageSaveError_Panic` — forces `Save` to fail on a `MessageEvent` and breaks on the yielded error. Before the fix it panics with the exact ticket stack (defer's yield at `runner.go:193` ← range body at `runner.go:234`). After the fix the consumer receives `ErrSessionSave` once and the deferred save logs through the "consumer stopped" branch.

[AI-1157]: https://redpandadata.atlassian.net/browse/AI-1157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -215,7 +215,14 @@ func (r *Runner) Run(
 			// Note: Agent already appended the message to sess.Messages, we just save it
 			if _, ok := evt.(agent.MessageEvent); ok {
 				if err := r.config.sessionStore.Save(ctx, sess); err != nil {
-					yield(nil, fmt.Errorf("%w: %w", agent.ErrSessionSave, err))
+					// Capture yield's bool so the defer's guard works if the
+					// consumer breaks here — otherwise the deferred save's
+					// yield panics with "range function continued iteration
+					// after function for loop body returned false".
+					if !yield(nil, fmt.Errorf("%w: %w", agent.ErrSessionSave, err)) {
+						consumerStopped = true
+					}
+
 					return
 				}
 			}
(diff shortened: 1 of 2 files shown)


weeco (approved):
Good fix, thanks

MikolajGagatekRP:
@claude review

PR #142, opened 2026-05-27
refactor(llm): sealed Part union

weeco:
Sealed `Part` interface for the public `llm/` API.

> The typed tool-schema change that originally rode along in this PR has been split out into #145 (draft) so this one stays focused on the sealed-`Part` refactor.

## Why
- The current `Part` struct has 6 fields, of which 4 are always `nil` for any given Part. Allocating one requires knowing which `Kind` discriminator picks which field — easy to get wrong, and the type system gives no help.
- Type switches over the discriminator are repetitive at every consumer.

## Before
```go
parts := []llm.Part{
    {Kind: llm.PartText, Text: "Hello"},
    {Kind: llm.PartToolRequest, ToolRequest: &llm.ToolRequest{
        ID: "call_1", Name: "search", Arguments: json.RawMessage(`{"q":"go"}`),
    }},
}

for _, p := range parts {
    switch p.Kind {
    case llm.PartText:
        fmt.Println(p.Text)
    case llm.PartToolRequest:
        fmt.Println(p.ToolRequest.Name)
    }
}
```

## After
```go
parts := []llm.Part{
    llm.NewTextPart("Hello"),
    llm.NewToolRequestPart("call_1", "search", json.RawMessage(`{"q":"go"}`)),
}

for _, p := range parts {
    switch p := p.(type) {
    case *llm.TextPart:
        fmt.Println(p.Text)
    case *llm.ToolRequestPart:
        fmt.Println(p.Name)
    }
}
```

Each Part owns exactly the fields it needs. Pointer receivers on the sealed marker prevent the footgun where both `TextPart{}` and `&TextPart{}` would satisfy the interface but only one matches a given type switch.

`ClonePart` and `CloneMessage` are added at session persistence boundaries so persisted state never shares mutable pointers with live state.

## Breaking
Pre-1.0, intentional. Consumers must:
- Construct Parts with `llm.New*Part()` helpers or `&llm.TextPart{...}` literals (value form no longer compiles)
- Type-switch on pointer-typed Parts (`case *llm.TextPart:`)

## Test plan
- [x] `go build`, `vet`, `task test:unit`, `task lint:new`
- [x] integration tests compile
- [x] examples build


weeco:
@claude Review the proposed changes. Are these changes worth breaking the API?

birdayz (approved):

PR #144, opened 2026-05-29
openai, anthropic: downgrade dynamic JSON Schema for strict tool input_schema

birdayz:
## What

Downgrade standard JSON Schema to each provider's accepted tool `input_schema`
subset in the OpenAI and Anthropic schema mappers. Adds a shared
`internal/jsonschema` package used by both. Google and openaicompat are
unchanged.

## Why

MCP servers expose full JSON Schema as tool parameters. OpenAI strict mode and
Anthropic reject several of those features outright (HTTP 400), so the model
never sees the tool. This broke real protoc-gen-go-mcp schemas containing
`bytes` fields, maps, and protobuf dynamic well-known types. The MCP ecosystem
(LangChain, Vercel AI SDK, OpenAI Agents SDK, RooCode) puts this adaptation on
the client, not the server; ai-sdk-go is that client.

Verified against the live providers:
- OpenAI strict rejects `format:"byte"`, `contentEncoding`, `propertyNames`, and
  any typeless node ("schema must have a 'type' key").
- Anthropic rejects typeless / open-ended nodes as invalid draft 2020-12.
- Gemini accepts all of it, hence no Google change.

## Implementation details

`internal/jsonschema`:
- `StripUnsupportedOpenAIKeywords`: removes `propertyNames`, `patternProperties`,
  `min/maxProperties`, `dependentRequired/Schemas`,
  `contentEncoding/MediaType/Schema`, and `format` values outside OpenAI's
  documented set (date-time, time, date, duration, email, hostname, ipv4, ipv6,
  uuid). Dropping `format:"byte"` folds a base64 hint into the description so the
  encoding signal survives the strip.
- `CollapseDynamicNodes`: rewrites typeless nodes, open objects
  (`additionalProperties: true`) and untyped arrays to a single JSON-encoded
  `string`. Targets exactly protobuf `Struct`/`Value`/`ListValue`/`Any.value` as
  they render in JSON Schema; leaves closed message objects, maps
  (`additionalProperties` is a schema), and typed scalars/arrays untouched.

`AdaptSchemaForOpenAI` = collapse + strip + the existing strict transform.
`AdaptSchemaForAnthropic` = collapse. Both deep-copy; inputs are never mutated.

The matching server-side piece (parsing a stringified dynamic WKT back to native
JSON on decode) lives in the consumer, protoc-gen-go-mcp, so the round trip is
closed there.

`openaicompat` is deliberately untouched: its endpoints range from strict to
fully permissive, so imposing the downgrade is the caller's decision.

Drive-by: the openaicompat Google compat test pinned `gemini-2.0-flash`, which is
retired and now 404s. Bumped to `gemini-3.5-flash`.

## References

protoc-gen-go-mcp PR: redpanda-data/protoc-gen-go-mcp#44

### DIFF
--- a/internal/jsonschema/downgrade.go
+++ b/internal/jsonschema/downgrade.go
@@ -1,0 +1,237 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package jsonschema holds provider-agnostic helpers for downgrading standard
+// JSON Schema (e.g. schemas exposed by MCP servers) into the narrower subsets
+// individual LLM providers accept as tool input_schema. MCP servers legitimately
+// publish full JSON Schema; per the MCP ecosystem the client is responsible for
+// adapting it per provider (the same way LangChain, the OpenAI Agents SDK and
+// the Vercel AI SDK do).
+package jsonschema
+
+// OpenAISupportedStringFormats is the set of JSON Schema string "format" values
+// OpenAI Structured Outputs / strict function calling accepts. Any other value
+// (e.g. "byte", "uri", "uri-reference") is rejected with a 400.
+// See https://platform.openai.com/docs/guides/structured-outputs.
+var OpenAISupportedStringFormats = map[string]bool{
+	"date-time": true,
+	"time":      true,
+	"date":      true,
+	"duration":  true,
+	"email":     true,
+	"hostname":  true,
+	"ipv4":      true,
+	"ipv6":      true,
+	"uuid":      true,
+}
+
+// StripUnsupportedOpenAIKeywords walks the schema and removes keywords OpenAI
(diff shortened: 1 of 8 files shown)


weeco:
Makes sense to me, though can we move the provider specific downgrade rules to the respective provider package instead of moving them all to the shared jsonschema package?

@claude Review

birdayz:
Done — pushed the restructure.

`internal/jsonschema/` now holds only provider-agnostic transforms:
- `CollapseDynamicNodes` (used by both OpenAI and Anthropic)
- `Walk` — exported as a reusable tree-walker
- `DeepCopy` — lifts the `deepCopyMap` that was copy-pasted in the openai and anthropic mappers (note 1)

OpenAI's allowlist, `stripUnsupportedKeywords` and the base64-hint helper now live in `providers/openai/` as unexported code and call `jsonschema.Walk`. The strip unit test moved with them. Anthropic and OpenAI both call `jsonschema.DeepCopy` now.

Also from the review notes:
- `CollapseDynamicNodes` is now a `Walk` pass instead of a second hand-rolled traversal (note 4) — collapsing a node clears its children, so `Walk` finds nothing to descend into; behaviour is identical, ~40 fewer lines.
- `Walk`'s doc comments that `fn` runs before recursion, so a delete also prunes the subtree (note 2).
- `appendBase64Hint` is a plain `if` on the string description now (note 3).

External behaviour is unchanged — the adapter integration tests and the live suite still pass.

birdayz:
@claude Review

The restructure addressing your feedback is pushed (commit a0f89f6c). Provider-specific rules now live in their provider packages; `internal/jsonschema/` holds only the cross-provider transforms (`CollapseDynamicNodes`, `Walk`, `DeepCopy`). Please take another look.

simon0191 (approved):

PR #147, opened 2026-06-09
feat(tool,runner): async tools — pause/resume, typed tool.Func, pre-1.0 API cleanups

weeco:
Async tools for the agent runtime: tools (and interceptors) can now **pause** an invocation — for external jobs, human approval, or user input — and the runner can **resume** it later, across process boundaries. Ships with the typed `tool.Func` path and a set of pre-1.0 API cleanups.

Builds on #142 (sealed `Part`).

## Why

Previously every tool had to return its final result inside one `Execute` call. That ruled out deployments, batch jobs, human approval gates, MCP elicitation — anything that takes minutes or needs a human in the middle. The SDK had wire format for "paused" but no mechanism to produce, persist, or resume a pause.

### Sync tool (the 90% path)

```go
type WeatherInput struct {
	City string `json:"city" jsonschema:"City name"`
}

type WeatherOutput struct {
	TempC float64 `json:"temp_c"`
}

var weather = tool.Must(tool.Func(
	tool.Spec{Name: "get_weather", Description: "Current weather for a city."},
	func(ctx context.Context, in WeatherInput) (tool.Result[WeatherOutput], error) {
		return tool.Done(WeatherOutput{TempC: 21.5}), nil
	},
))
```

Schema is inferred from `WeatherInput`. No JSON plumbing, no `Definition()` boilerplate.

### Async tool (external job)

```go
var deploy = tool.Must(tool.Func(
	tool.Spec{
		Name:        "deploy",
		Description: "Trigger a production deployment.",
		Async:       tool.AsyncExternalResult(), // model is told not to re-call while pending
	},
	func(ctx context.Context, in DeployInput) (tool.Result[DeployStatus], error) {
		jobID := startDeployment(in)

		return tool.Pending(
			DeployStatus{ID: jobID, Status: "queued"}, // placeholder the model sees meanwhile
			tool.WithCorrelationID(jobID),
			tool.WithAwaitTimeout(30*time.Minute),
		), nil
	},
))
```

The invocation ends with `FinishReasonPaused`; the pending call is persisted on the session (`session.PendingToolCall`, kvstore included) and surfaced on `InvocationEndEvent.PendingCalls`.

### Resuming (e.g. from a webhook handler)

```go
stream, err := r.Resume(ctx, actorID, sessionID,
	runner.Resumption{CallID: callID, Output: finalResultJSON})
if err != nil {
	// eager: ErrResumeConflict, ErrPendingCallNotFound, ErrInvalidResumePayload
}

for evt, err := range stream { // agent continuation (model reacts to the result)
	...
}
```

`Resume` mutates **eagerly** — by the time it returns, the session is updated and saved; dropping the stream skips only the model continuation. It is idempotent for at-least-once delivery: duplicate submissions with the same payload are acknowledged, different payloads conflict (Stripe-style). `Runner.Progress` records non-terminal updates; `Runner.Cancel` aborts.

### Human-in-the-loop approval (interceptor)

```go
func (g *Gate) InterceptToolExecution(
	ctx context.Context, info *agent.ToolCallInfo, next agent.ToolExecutionNext,
) (tool.Execution, error) {
	if info.Resume == nil { // first entry: pause instead of executing
		return tool.Execution{
			Output: json.RawMessage(`{"status":"awaiting_approval"}`),
			Await:  &tool.Await{Reason: tool.AwaitReasonApproval, Message: "Approve " + info.Req.Name + "?"},
		}, nil
	}

	decision := info.Resume
	info.Resume = nil // consume the decision; the tool runs fresh below

	if decision.Error != "" {
		return tool.Execution{}, errors.New(decision.Error) // denied → model-visible tool error
	}

	return next(ctx, info) // approved → tool executes exactly once
}
```

Re-entry goes through the interceptor chain, so denial means the tool **never runs**. Chained pauses work: approve → tool itself returns `Pending` → resume again with the external result (covered by an end-to-end test).

## Breaking changes (pre-1.0, intentional)

- `tool.Tool` is now `Name()/Description()/InputSchema()` + `Execute(ctx, Call) (Execution, error)`; errors returned from `Execute` become model-visible tool errors.
- `tool.Registry` is a concrete struct; `Execute`/`ExecuteAll` removed (use `Run(...).Response()`); `NewRegistry()` is variadic. MCP accepts a narrow `mcp.ToolRegistry` interface.
- `Runner.Resume/Cancel` return `(iter.Seq2[agent.Event, error], error)`; `Progress` returns `(agent.ToolProgressEvent, error)`.
- `agent.ToolInterceptor` returns `tool.Execution`; `ToolCallInfo` gained `Resume`.
- Removed: `Call.Args`, `Await.ExpiresAt` (use `Timeout`), `Execution/Action/Config.Metadata` (write-only), `Spec.OutputSchema` (advisory-only), `tool.PendingReentry` (the typed path never re-runs the function; implement `Tool` directly for re-entry logic), coalescing fields. `Await.Resume` may be left empty (defaults from `Reason`).
- `kvstore` persists `PendingToolCalls`/`ResumeReceipts` (new proto fields 4/5, backward compatible).

## Known gaps (deliberate, follow-ups)

- Crash recovery re-executes incomplete tools and cannot re-bind pauses produced during recovery (warned loudly at runtime).
- A2A/Vercel adapters surface pauses via `InvocationEndEvent.PendingCalls` but don't yet map `ToolProgressEvent`/`ToolArtifactEvent` to their wire formats.
- Session locks are process-local; distributed deployments need an external advisory lock (documented on `session.Store`).


weeco:
@claude Review this, specifically whether API and usage for users is good

PR #149, opened 2026-06-10
fix(providers): tighten fable 5 model support

malinskibeniamin:
## Summary

Follow-up to #146.

- removes unsupported sampling params (`temperature`, `top_p`, `top_k`) from first-party Anthropic `claude-fable-5` so callers fail fast client-side instead of getting API 400s
- adds Fable 5 option coverage for sampling rejection, thinking-budget rejection, speed rejection, and supported effort levels
- adds Bedrock Fable 5 catalog entries for documented `global.`, `us.`, and `eu.` inference profiles
- exposes `requires_provider_data_sharing=true` in model discovery metadata for Bedrock Fable 5 entries
- adds Bedrock lookup, region-prefix, sampling-param, pricing, discovery-metadata, and conformance coverage for the new entries
- adds a dedicated Bedrock Fable 5 integration that hits Bedrock and either generates successfully or skips on AWS's explicit Fable access gates (provider data sharing / account access)

Docs checked:

- Anthropic Fable 5 migration notes: https://platform.claude.com/docs/en/about-claude/models/migration-guide
- AWS Fable 5 Bedrock model card: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-fable-5.html
- AWS Bedrock data retention docs: https://docs.aws.amazon.com/bedrock/latest/userguide/data-retention.html

## Test plan

- `go test ./providers/anthropic/... -short`
- `go test ./providers/bedrock/... -short`
- `go test ./providers/bedrock -run 'TestModelRequiresProviderDataSharing|TestIsProviderDataSharingGate|TestIsFableAccessGate|TestBedrockFable5Invocation_Integration|TestBedrockConformance_Integration|TestModelsDiscovery_ProviderDataSharingMetadata' -count=1`
- `go test ./llm -count=1`
- `task test:unit`
- `task lint:new`
- `task license:check`

### DIFF
--- a/providers/anthropic/models.go
+++ b/providers/anthropic/models.go
@@ -115,7 +115,7 @@ var supportedModels = map[string]ModelDefinition{
 			MaxOutputTokens:  128000,  // 128K output tokens
 			// Fable 5 rejects thinking.type.enabled — thinking budget is not user-controllable.
 			// Use adaptive thinking + effort to bias reasoning depth. No fast mode, so no "speed".
-			SupportedParams:   []string{"temperature", "top_p", "top_k", "max_tokens", "effort"},
+			SupportedParams:   []string{"max_tokens", "effort"},
 			MutuallyExclusive: [][]string{},
 		},
 		SupportedEfforts: []Effort{EffortLow, EffortMedium, EffortHigh, EffortXHigh, EffortMax},
--- a/providers/bedrock/bedrock_conformance_test.go
+++ b/providers/bedrock/bedrock_conformance_test.go
@@ -16,9 +16,11 @@ package bedrock_test
 
 import (
 	"context"
+	"errors"
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/redpanda-data/ai-sdk-go/internal/testsuite"
 	"github.com/redpanda-data/ai-sdk-go/llm"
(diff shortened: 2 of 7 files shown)


providers/anthropic/models.go
```diff
@@ -115,7 +115,7 @@ var supportedModels = map[string]ModelDefinition{
 			MaxOutputTokens:  128000,  // 128K output tokens
 			// Fable 5 rejects thinking.type.enabled — thinking budget is not user-controllable.
 			// Use adaptive thinking + effort to bias reasoning depth. No fast mode, so no "speed".
-			SupportedParams:   []string{"temperature", "top_p", "top_k", "max_tokens", "effort"},
+			SupportedParams:   []string{"max_tokens", "effort"},
```
malinskibeniamin, line 118: there is no temperature setting for fable 5

birdayz:
does it run actual tests vs. bedrock ? it's a bit of a pain to test, want to be sure it's actually wired up correctly

providers/bedrock/bedrock_conformance_test.go
```diff
@@ -101,6 +101,10 @@ func (f *BedrockFixture) Models() []llm.ModelDiscoveryInfo {
 	filtered := make([]llm.ModelDiscoveryInfo, 0, len(all))
 
 	for _, m := range all {
+		if requiresProviderDataSharing(m.Name) && os.Getenv("BEDROCK_INCLUDE_PROVIDER_DATA_SHARING_MODELS") != "1" {
```
birdayz: this env..is a smell. why does the unit test need to look at the env. either we test against this model, or we don't. and in the AIGW, we can just make it visible that it requires data sharing

malinskibeniamin:
@claude review

birdayz (approved):

PR #153, opened 2026-06-17
feat(agenttool): share parent session id with in-process sub-agents

alenkacz:
## What

In-process sub-agents (`agenttool`) minted a fresh `agent-tool-<name>-<nano>` session id per call, so the OTel conversation id differed from the parent and the two appeared as **separate conversations** in session-oriented observability (e.g. Langfuse) — even though the trace tree is already connected via `ctx` propagation.

This makes an `agenttool` sub-agent **share its parent's session id** when invoked as part of a parent agent's turn, so the two group into one conversation for tracing/reconstruction. Context stays fully isolated.

## How

- **`agent`**: new `ContextWithInvocation` / `InvocationFromContext` — the missing channel that lets a tool see its parent invocation. `llmagent.executeTools` injects the invocation into `ctx` before executing tools (additive; tools that don't read it are unaffected).
- **`agenttool`**: when a parent invocation is in `ctx`, the sub-agent shares the parent's session id and stamps linkage (`parent_invocation_id`, `agent_path`) into session metadata. Falls back to the previous minted id when there is no parent invocation.
- **`store/session`**: canonical `Metadata*` linkage key constants (single source of truth).
- **`plugins/otel`**: emits `redpanda.agent.parent_invocation_id` / `redpanda.agent.path` span attributes on the invocation span.

## Isolation & safety

- Isolation is preserved: the sub-agent's `Messages` are still built **fresh in memory and never loaded**, so it cannot observe the parent's history — sharing the id only affects grouping, not context.
- The sub-agent path never persists (no Runner, no store on `llmagent`), so the shared id is safe as a non-storage-key. A code comment records the guard: if persistence is ever added on this path, key on the unique invocation id / `agent_path`, **not** the shared session id.

## Linkage surface (maps to Claude's model)

| Concern | Field | Claude analog |
|---|---|---|
| Grouping | shared `session.ID` | `sessionId` |
| Per-sub-agent identity | child `InvocationID` | `agentId` |
| Parent back-reference | `parent_invocation_id` | `parentUuid` / `toolUseId` |

(An earlier `is_sidechain` boolean was dropped — it was Claude-internal jargon and redundant with `parent_invocation_id`, whose presence is itself the sub-agent signal.)

## Tests

`agent/context_test.go` and `tool/agenttool/shared_session_test.go` cover: round-trip ctx helpers, shared id, **context isolation despite shared id**, no-parent fallback, and nil-session fallback. Existing `agent`/`llmagent`/`otel`/`runner` suites pass; `go vet`, `golangci-lint`, `gofmt` clean.

### DIFF
--- a/agent/context.go
+++ b/agent/context.go
@@ -1,0 +1,43 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package agent
+
+import "context"
+
+// invocationCtxKey is the unexported key under which the active invocation
+// metadata is stored in a context.
+type invocationCtxKey struct{}
+
+// ContextWithInvocation returns a copy of ctx carrying the given invocation
+// metadata. Agents put their invocation into the context before executing tools
+// so that tools (notably agenttool) can access the calling agent's invocation —
+// for example to share the parent's session id with an in-process sub-agent.
+//
+// A nil invocation returns ctx unchanged.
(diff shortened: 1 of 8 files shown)


birdayz:
**Codex finished review in 6m26s** · `63d26aac`

---

The changes compile and the relevant/full test suites pass. I did not identify any introduced correctness issues that warrant an inline finding.

---

`29 tool-calls` · Generated by Codex 0.139.0 on gpt-5.5 (xhigh) and [codex-review](https://github.com/birdayz/skills/tree/master/codex-review)

PR #157, opened 2026-06-19
fix: make prompt caching work (deterministic tool order + default-on Anthropic/Bedrock caching)

birdayz:
## What

Make prompt caching actually work for agents built on this SDK. Two changes,
each fixing a class of bug that silently destroys the upstream cache:

1. **Deterministic tool ordering.** `tool.Registry.List()` and the MCP client's
   `ListTools()` now return tool definitions sorted by name instead of in Go's
   randomized map-iteration order.
2. **Caching on by default for the Anthropic-backed providers.** The Anthropic
   and Bedrock providers now enable prompt caching by default. A new
   `WithCachingDisabled()` option turns it off; the existing no-arg
   `WithCaching()` stays valid (now a no-op).

No public signatures change. The only behavioral change a caller can observe is
that requests are now deterministic and Anthropic/Bedrock requests carry
`cache_control` / `CachePoint` markers unless explicitly disabled.

## Why

Caching is effectively dead for managed agents today, for two independent
reasons.

**Nondeterministic tool order busts the cache prefix on every provider.**
`registry.List()` ranged the internal `map[string]*registeredTool` with no sort:

```go
for _, registered := range r.tools {        // map range: randomized every call
    definitions = append(definitions, registered.tool.Definition())
}
```

`llmagent` calls `List()` fresh on every turn (`agent/llmagent/llmagent.go:251`,
`:670`) and every provider's request mapper serializes `req.Tools` in slice
order (verified for anthropic, openai, openaicompat, google, bedrock — none
re-sort). Go randomizes map iteration per range, so the tools array goes to the
wire in a different order on every request. For Anthropic the tools block
precedes system + messages and the breakpoint sits on the last system block, so
the cached prefix is `[tools]+[system]`; reorder the tools and the prefix never
matches. The same byte-level change defeats OpenAI's and Google's automatic
prefix caching and busts the prefix that Bedrock's explicit CachePoint markers
cover. An agent with several MCP servers has many tools, so this reorders
constantly. A single test that calls `List()` 50 times on a 16-tool registry
observed 14 distinct orderings.

The MCP client's `ListTools()` (`tool/mcp/tools.go:50`) has the identical
map-range bug. The agent path doesn't hit it (MCP tools are registered into the
registry one by one, so the registry sort covers that path), but `ListTools()`
is a public API whose docstring explicitly says the result is "Passed to an LLM
for tool calling." Anyone building `llm.Request.Tools` from it directly gets the
same cache-busting nondeterminism, and the registry sort does not reach it.

**Caching is opt-in and nobody opts in.** Anthropic's `EnableCaching` defaulted
`false`; `cache_control` markers are written only when it is true; the only way
to flip it was `WithCaching()`. Managed-agent construction never calls it, so
zero markers were ever emitted. Bedrock (which runs Claude on AWS) has the exact
same shape: `enableCaching` defaults `false`, only `bedrock.WithCaching()` flips
it. Bedrock does **not** do automatic prefix caching — the explicit CachePoint
markers are the only way to cache there at all, so default-off meant Bedrock
agents never cached, full stop. For an agent hammering a stable system prompt +
toolset, cache reads are ~90% cheaper; the only cost is a one-time ~25% write
premium, and below the provider's token minimum the marker is simply ignored.
Opt-in just means everyone forgets — which is the bug.

OpenAI, OpenAICompat and Google need no flag change: they cache automatically
server-side with no markers. Their only exposure was the tool-ordering bug,
which the sort fixes.

## Implementation details

This change is gated by unit tests that are red before the fix and green after,
committed in that order so the regression is auditable. The tests are pure unit
tests (no network, no API key) so they gate CI — unlike the existing
integration-only cache tests.

**Tool ordering.** `registry.List()` sorts the assembled slice by
`Definition().Name` before returning; `mcp.clientImpl.ListTools()` does the
same. Tool names are unique (the registry rejects duplicates; MCP namespaces by
server ID), so the sort is a total order and fully deterministic. This is the
single source of truth — every provider consumes the now-sorted `req.Tools`
downstream, so one sort fixes every provider. The audit confirmed no provider
mapper re-sorts or re-randomizes, and that schema mappers build their `required`
arrays deterministically (`encoding/json` already marshals map keys sorted).

**Default-on caching.** `anthropic.NewProvider` and `bedrock.NewProvider` set
the caching flag to true before applying options. `WithCaching()` keeps setting
it true (harmless no-op, kept so existing callers and tests compile unchanged —
adding a bool parameter would be a breaking change). A new
`WithCachingDisabled()` option sets it false for callers who genuinely want to
opt out. The marker-writing logic is unchanged: Anthropic marks the last system
block and the last message; Bedrock inserts a `CachePoint` after the last system
block and the last message content.

**Backward compatibility.** No exported signature changes. `WithCaching()`
remains callable with no arguments. The behavioral delta is deterministic
request bodies and default-on Anthropic/Bedrock markers, both of which are
strictly cache-positive and free below the token minimum.

**What this does not fix.** Deterministic tool order is the SDK-controlled
cache-buster, but it is not the only way a prefix can change turn to turn, and
this PR should not be read as "caching can never miss":

- A **volatile system prompt** defeats everything. `llmagent` calls the
  caller-supplied system-prompt provider every turn; if it interpolates a
  timestamp, date, or session ID, the system breakpoint rewrites every turn and
  tool order is irrelevant. That is the caller's responsibility.
- The valuable breakpoint is the one on the last **system** block (it caches the
  stable tools+system prefix). The last-message marker is incidental — in a
  multi-turn loop it mostly pays write premiums — and is left unchanged here.
- Anthropic's cache lookback is the last 20 content blocks; a single turn with
  more than 20 tool_use/tool_result blocks can push the prior breakpoint out of
  the window and miss. Not introduced here, but default-on makes it relevant to
  tool-heavy agents. Tracked as a follow-up.

## References

- Follow-up (separate, out of scope here): `cloudv2` bumps its pin to pick this
  up. With default-on caching, `apps/ai-agent` needs no caching call-site change
  — `go get github.com/redpanda-data/ai-sdk-go@main && go mod tidy` is the whole
  diff.

### DIFF
--- a/providers/anthropic/cache_default_test.go
+++ b/providers/anthropic/cache_default_test.go
@@ -1,0 +1,106 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package anthropic
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/anthropics/anthropic-sdk-go"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/redpanda-data/ai-sdk-go/llm"
+)
+
+// TestDefaultProviderEnablesCaching pins the default: a provider built the way
+// every consumer builds it (no caching option) must have caching ON. Prompt
+// caching being opt-in means everyone forgets to opt in, which is exactly the
+// regression this guards against.
+func TestDefaultProviderEnablesCaching(t *testing.T) {
+	t.Parallel()
+
+	p, err := NewProvider("test-key")
+	require.NoError(t, err)
+
+	assert.True(t, p.EnableCaching,
+		"caching must default to ON; the SDK's consumers never call WithCaching()")
+}
+
+// TestDefaultCachingEmitsCacheControlMarkers proves the default flows all the
+// way through: a model built from a default provider must emit cache_control
+// markers on the cacheable prefix (last system block) and on the last message,
+// without anyone calling WithCaching().
+func TestDefaultCachingEmitsCacheControlMarkers(t *testing.T) {
+	t.Parallel()
+
+	p, err := NewProvider("test-key")
+	require.NoError(t, err)
+
+	model, err := p.NewModel(ModelClaudeSonnet45)
+	require.NoError(t, err)
+
+	m, ok := model.(*Model)
+	require.True(t, ok, "NewModel must return *Model")
+
+	req := &llm.Request{
+		Messages: []llm.Message{
+			{
+				Role: llm.RoleSystem,
+				Content: []llm.Part{
+					// Realistic system prompt; size is irrelevant to marker
+					// emission (Anthropic ignores markers below its token
+					// minimum server-side, the SDK always writes them).
+					llm.NewTextPart("You are a helpful assistant. " + strings.Repeat("context ", 200)),
+				},
+			},
+			{
+				Role:    llm.RoleUser,
+				Content: []llm.Part{llm.NewTextPart("Hello")},
+			},
+		},
+	}
+
+	apiReq, err := m.requestMapper.ToProvider(req)
+	require.NoError(t, err)
+
+	require.NotEmpty(t, apiReq.System, "system blocks must be present")
+	lastSystem := apiReq.System[len(apiReq.System)-1]
+	assert.True(t, hasCacheMarker(lastSystem.CacheControl),
+		"last system block must carry a cache_control marker by default")
+
+	require.NotEmpty(t, apiReq.Messages, "messages must be present")
+	assert.True(t, lastMessageHasCacheMarker(apiReq.Messages[len(apiReq.Messages)-1]),
+		"last message must carry a cache_control marker by default")
+}
+
+// hasCacheMarker reports whether a cache_control value has been set. The zero
+// value leaves Type empty; NewBetaCacheControlEphemeralParam sets it to
+// "ephemeral".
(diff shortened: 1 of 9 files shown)


birdayz:
**Codex finished review in 1m29s** · `4f248fb2`

---

The patch adds tests whose expectations are not implemented by the current code: caching still defaults off for Anthropic and Bedrock, and tool listing still returns map iteration order. These changes would leave the affected package tests failing or flaky.

Full review comments:

- [P1] Enable Anthropic caching before adding default-on assertions — providers/anthropic/cache_default_test.go:38-39
  When `NewProvider("test-key")` is used without `WithCaching()`, the current provider leaves `EnableCaching` at its bool zero value and `NewModel` copies that into the request mapper. This new assertion, and the marker checks below, therefore fail under `go test ./providers/anthropic` until the default is actually changed or the test opts in.

- [P1] Set Bedrock caching default before asserting it — providers/bedrock/cache_default_test.go:40-41
  With only `WithAWSConfig`, `NewProvider` stores `cfg.caching`, which defaults false, into `p.enableCaching`; `NewModel` then builds a mapper that never inserts `CachePoint`s. These new default-on tests fail in `go test ./providers/bedrock` unless the Bedrock provider default is changed or the test passes `WithCaching()`.

- [P1] Sort registry results before requiring sorted output — tool/registry_test.go:59-60
  `Registry.List()` still appends definitions by ranging over `r.tools`, so for any registry with multiple tools its order is Go map iteration order, not the sorted `want` built here. This assertion will fail or be probabilistic in `go test ./tool` until `List()` sorts the definitions by name.

- [P1] Sort MCP tool definitions before requiring sorted output — tool/mcp/tools_list_test.go:62-62
  `clientImpl.ListTools()` currently ranges over `c.tools` and returns that map order verbatim. For a started client with multiple tools like this test constructs, the output is not guaranteed to match `sort.Strings(want)`, so `go test ./tool/mcp` will fail or flap until the method sorts by definition name.

---

`14 tool-calls` · Generated by Codex 0.139.0 on gpt-5.5 (xhigh) and [codex-review](https://github.com/birdayz/skills/tree/master/codex-review)

birdayz:
**Codex finished review in 7m59s** · `ce08edce`

---

The changed code compiles and the relevant package tests pass. I did not find any discrete regressions introduced by the caching defaults or deterministic tool ordering changes.

---

`32 tool-calls` · Generated by Codex 0.139.0 on gpt-5.5 (xhigh) and [codex-review](https://github.com/birdayz/skills/tree/master/codex-review)

alenkacz (approved):

PR #162, opened 2026-07-02
fix: don't emit empty-content assistant messages on replay

MikolajGagatekRP:
## What

Fixes a 400 from Anthropic's Messages API (`messages.N.content: Field required`) that can happen when replaying session history after a `max_tokens`-truncated assistant turn.

Two layers:
- **Authoritative** — `providers/anthropic/request_mapper.go` (`mapAssistantMessage`): if an assistant message maps to empty API content, substitute a single minimal text block.
- **Defense-in-depth** — `agent/llmagent/llmagent.go`: skip persisting an assistant turn whose content is empty, so it never enters the session store in the first place.
- `providers/openaicompat/request_mapper.go`: comment documenting the existing equivalent guard (empty assistant turns are dropped by construction).

## Why

An assistant turn truncated at `max_tokens` (`stop_reason=max_tokens` → `FinishReasonLength`) can be finalized with an **empty content slice**: if the model goes straight to tool calls and the final `tool_use` block is cut off, the streaming finalizer (`providers/anthropic/model.go`) drops the unparseable partial block. If every block was such a dropped partial, the finalized message has zero parts.

That message was:
1. persisted unconditionally to the session (`llmagent.go`), then
2. replayed on the next request, where `mapAssistantMessage` produced an assistant message with an empty content array →
3. Anthropic rejects it: `400 messages.N.content: Field required`.

The existing `detectIncompleteToolCalls` recovery misses this because it only fires when the prior assistant message *has* tool requests — here they were dropped.

## How

**Substitute vs drop (mapper):** Chose substitute. On replay the poisoned turn sits mid-history as `user, assistant(empty), user`. Dropping it yields `user, user`, which breaks Anthropic's required user/assistant alternation. Substituting a minimal text block keeps the sequence valid; a truncated turn carries no recoverable content, so the placeholder loses nothing. (The replayed turn is never the final message, so the substituted block only needs to be a non-empty content array — which resolves the `content: Field required` error.)

**Persist guard (`len > 0`):** A legitimately truncated tool-use turn keeps its completed `tool_use` parts (len > 0) and is unaffected — only genuinely content-less turns are skipped. `FinishReason` is read from the response, not the session, so `FinishReasonLength` still propagates.

**openaicompat:** shape differs — the mapper groups parts and only emits an assistant message when text or tool calls are present, so an empty-content turn is already never emitted as an invalid content-less message. The Chat Completion API tolerates the resulting sequence, so no placeholder is needed. Added a clarifying comment; behavior unchanged (an existing test locks in that reasoning-only assistant turns are dropped).

## Tests

- `providers/anthropic/empty_content_test.go`
  - `TestMapAssistantMessage_EmptyContentIsRepaired` — mapper-level: history with an empty-content assistant message must not map to any empty-content API message.
  - `TestStreaming_MaxTokensAllBlocksDropped_ReplayStaysValid` — end-to-end: stream a `max_tokens` cut whose *only* block is a partial `tool_use`; assert the finalized message has empty content + `FinishReasonLength`, then map that history back and assert no empty-content API message.
- `agent/llmagent/empty_content_persist_test.go` — an empty-content `FinishReasonLength` turn is not persisted, while the truncation signal still surfaces.
- `providers/openaicompat/empty_content_test.go` — parity: empty assistant turn is dropped, not emitted as an invalid message.

Verified red→green: the two anthropic tests and the llmagent test fail on the unfixed code (with the exact `content: Field required` symptom) and pass after the fix; `go build`, `go vet`, and `go test ./providers/anthropic/... ./providers/openaicompat/... ./agent/...` are all green.

### DIFF
--- a/agent/llmagent/empty_content_persist_test.go
+++ b/agent/llmagent/empty_content_persist_test.go
@@ -1,0 +1,75 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package llmagent_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/redpanda-data/ai-sdk-go/agent"
+	"github.com/redpanda-data/ai-sdk-go/agent/llmagent"
+	"github.com/redpanda-data/ai-sdk-go/llm"
+	"github.com/redpanda-data/ai-sdk-go/llm/fakellm"
+	"github.com/redpanda-data/ai-sdk-go/store/session"
+)
+
+// TestRun_EmptyContentTurnIsNotPersisted is the defense-in-depth regression for
+// the empty-content replay bug. When a max_tokens cut yields an assistant
+// message with zero content parts (its only block was a partial tool_use the
+// provider dropped), the agent loop must NOT append that turn to the session.
+// Persisting it poisons the store: the next request replays a content-less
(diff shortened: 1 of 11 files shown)


MikolajGagatekRP:
@claude review

MikolajGagatekRP:
@claude review

MikolajGagatekRP:
@claude review

birdayz (approved):

MikolajGagatekRP:
@claude review

PR #163, opened 2026-07-02
session: add List to the Store interface

birdayz:
## What

Add `List` to the `session.Store` interface, plus the `Summary` / `ListRequest` / `ListResponse` types, an `ErrListNotSupported` sentinel, and an `UpdatedAt` field on `session.State`.

## Why

The interface only exposed `Load` / `Save` / `Delete` keyed by ID, so any consumer that needs to enumerate an agent's sessions has to reach around it. Listing is a normal session-store capability - adk-go's `session.Service` has it - and it belongs on the interface. The immediate consumer is cloudv2's ai-agent `SessionService`, which backs an inspector session picker.

## Implementation details

`List` returns lightweight `Summary` values (id, metadata, updated_at) with no message payload, ordered by `updated_at` descending then `id` ascending. The page token is opaque, so a durable backend can use keyset paging and the in-memory one an offset - neither leaks into the contract.

`State` gains `UpdatedAt`: storage-managed and output-only. `Save` ignores whatever the caller put there and the store stamps the write time; `Load` and `List` populate it; `Clone` copies it.

`InMemoryStore` implements `List` over its cache (`Range` + sort) and stamps `UpdatedAt` on `Save`. The kvstore returns `ErrListNotSupported` - it is keyed by session ID and cannot enumerate, and it is slated for deprecation.

Adding a method to `Store` is breaking for every implementer; the only other one in-tree is a runner test mock, which gains `List`.

## References

Consumer: cloudv2 ai-agent `redpanda.adp.managedagent.v1.SessionService`.

### DIFF
--- a/runner/runner_test.go
+++ b/runner/runner_test.go
@@ -671,6 +671,7 @@ type mockSessionStore struct {
 	loadFunc   func(context.Context, string) (*session.State, error)
 	saveFunc   func(context.Context, *session.State) error
 	deleteFunc func(context.Context, string) error
+	listFunc   func(context.Context, *session.ListRequest) (*session.ListResponse, error)
 }
 
 func (m *mockSessionStore) Load(ctx context.Context, sessionID string) (*session.State, error) {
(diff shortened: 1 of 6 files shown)


birdayz:
@claude plz review

weeco (approved):

PR #164, opened 2026-07-03
feat(vercelaisdk): agent-backed UI message stream handler with server-side sessions

birdayz:
## What

`uimessagestream.Handler`: serve a chat resource over the Vercel AI SDK UI Message Stream protocol (what `useChat` speaks), backed by a full `agent.Agent` and a `session.Store` with server-side sessions. Replaces both the old model-level `Handler` and the interim client-authoritative design.

```
POST   /{$}   run a turn (SSE UI message stream)
GET    /{id}  history as UI messages, for useChat resume
DELETE /{id}  delete the chat
GET    /{$}   list chats
```

## Why

Two seams were wrong.

The old handler bound at the `llm.Model` layer and ran its own agentic tool loop. The system prompt, tool registry (MCP + subagents-as-tools), interceptor chain (OTel/transcripts), and the loop all live in `agent.Agent`, above the model — pointing the model-level handler at an agent would run a second loop and re-plumb telemetry.

And history belongs to the server. The Vercel AI SDK's canonical persistence pattern is load-chat-by-id, append the one posted message, run, save on finish; the client sends only `{id, message}`. Trusting a browser-posted history means a client can forge assistant turns and tool results, and every deployment ends up re-implementing persistence anyway. Sessions persist as `llm.Message` in the existing `session.Store` — the same store the A2A adapter and runner use — so one conversation record serves every protocol.

## Implementation details

- **Persistence is a wrapper, not a fork.** `persistingAgent` implements `agent.Agent` in front of the unchanged `StreamAgent`: save before the run, after every completed assistant message, and on exit — always with `context.WithoutCancel`, so a client disconnect cannot lose a completed turn. Not the runner: its append-one-message shape cannot express regenerate, and its deferred save dies with the request context.
- **Regenerate** (`trigger: regenerate-message`) truncates to the last genuine user message and re-runs without appending. Tool results are persisted as user-role messages and must not count as truncation targets. Posted history is never trusted; only the last message of an untrimmed body is used.
- **History projection.** `GET /{id}` renders stored model messages as UI messages: contiguous assistant + tool-result runs merge into one assistant message with `step-start` parts — the exact shape `useChat` assembles from a stream. Dangling tool requests close as `output-error`; all error text routes through the same `WithOnError` mapper as the live stream, so a page reload cannot leak what the stream sanitized.
- **Tenancy.** `WithSessionKey` derives the storage key from the authenticated request; every route resolves through it. Configuring it disables the list endpoint (`session.Store.List` cannot be tenant-scoped) — multi-tenant list APIs stay app-level.
- **Concurrency.** POSTs to one chat serialize on a refcounted keyed lock (in-process; multi-replica deployments serialize externally).
- **Dynamic tools.** Agent tools are runtime-discovered, so tool chunks carry `dynamic: true` (`DynamicToolUIPart`).
- **Defended stream grammar.** Steps open lazily on first content; every terminal path — finish, error, abort, cancellation — closes open text spans, steps, and unresolved tool calls before the terminator, and at most one `error` chunk is emitted per stream. `max_turns`/`input_required` emit `error` then `finish{error}`. Keep-alive SSE pings cover slow tool calls; token usage rides `messageMetadata` on `finish`.

## Verification

Conformance suite (`task test:conformance`) drives the real `ai@7.0.6` client (`DefaultChatTransport` + `readUIMessageStream`) against `Handler` over scripted fake agents: 16 tests including multi-turn accumulation over the trimmed transport, regenerate, history resume, delete/list, and the untrusted-posted-history rule. Additionally verified end-to-end with a live OpenAI model and an independent integration review against the README.

`examples/vercelaisdk_chat` is a runnable full-stack demo (video below): a Go server with a real OpenAI model and one tool, and a vanilla `useChat` React client — streaming tool calls, server-side memory over the trimmed transport, resume after reload, regenerate, chat list/delete. `web/src/App.tsx` is the entire client integration.

## References

Companion cloudv2 RFC + wiring: redpanda-data/cloudv2#27697

https://github.com/user-attachments/assets/1c51fa44-659e-46a7-bd32-358252f2412e


birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=2 · P2=1 · P3=0

The new AgentHandler can execute forged client-supplied incomplete tool calls and exposes raw agent tool errors, which are blocking security/correctness issues for a browser-facing adapter. It also misreports length-limited completions as errors.

Full review comments:

- [P1] Reject incomplete tool calls from client history — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:107-107
  Because this handler seeds an LLMAgent session directly from client-supplied UI history, LLMAgent.Run's crash-recovery path will treat a history ending with an assistant tool part in state `input-available` followed by the new user message as an incomplete persisted call and execute that tool before asking the model. For browser-facing AgentHandler this lets the caller forge an assistant message to force any registered MCP/subagent tool; drop/reject incomplete tool requests or disable recovery before constructing the session.

- [P1] Sanitize agent tool errors before streaming — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:203-204
  When an agent tool fails, LLMAgent/tool registry stores the real `err.Error()` in `ToolResponsePart.Result`; this branch writes that raw payload to the client as `errorText`. With the default AgentHandler options callers expect the same sanitized `onError` behavior as `Handler`'s executor path, but failed MCP/subagent tools can leak server-side details in `tool-output-error` chunks.

- [P2] Map length finishes to length — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:403-404
  When the agent stops because the model hit a length/context limit, the UI Message Stream protocol has a `length` finishReason and the existing model handler uses it. Mapping this to an `error` chunk makes useChat call its error path and loses the partial-completion semantics whenever `llmagent` maps `llm.FinishReasonLength` to `agent.FinishReasonLength`.

Script done on 2026-07-03 19:41:52+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Implementation review (torvalds, Vercel-AI-SDK-maintainer, codex) + fixes

Reviewed by three independent lenses. The Vercel-maintainer pass confirmed **no blocking protocol issues** — every emitted chunk validates against ai@7.0.6's `z.strictObject` union and the ordering is what a real `useChat` client tolerates (verified against `process-ui-message-stream.ts`). The real fixables, now applied in the latest commit:

**P1**
- **Incomplete/forged tool calls (codex + torvalds).** `reconstructAssistant` emitted a bare `tool_use` for a client-supplied tool part in state `input-available` — no paired `tool_result`. That both trips Anthropic's call/result pairing (400) and, worse, lets a browser forge an incomplete assistant tool call that an agent's crash-recovery path would execute before consulting the model. Fixed: the tool-request part is now emitted only in the result-bearing states; unresolved calls are dropped.
- **Tool-error leak (codex).** `tool-output-error` streamed the raw `err.Error()` from the tool registry. Now routed through `onError` (sanitized by default), matching the model-level executor.

**P2 / should-fix**
- **Length finish (codex + vercel).** `FinishReasonLength` now maps to the UI `length` reason (partial completion, no error chunk) instead of `error`.
- **Errored finish text (torvalds).** `FinishReasonError` now always emits a sanitized error chunk, matching the iterator-error / incomplete-run paths.

**Nits**
- `messageMetadata.usage` keys → camelCase (SDK `LanguageModelUsage` convention).
- Confirmed-correct and left as-is: lazy step open / balanced terminals, `tool-output-available` omitting `toolName`, keep-alive SSE comment, abort semantics, dynamic-tool typing.

Tests added for the dropped incomplete call, tool-error sanitization, and the length finish. Package tests + lint green.

birdayz:
## Update: removed the model-level handler

Two follow-up commits since the review:

1. **Review fixes** — dropped incomplete/forged tool calls in history reconstruction (Anthropic 400 + injection vector), sanitized tool errors through onError, mapped length→`length` (not error), always emit error text on an errored finish, camelCase usage metadata. Tests added.
2. **Dropped the model-level handler entirely.** It shipped its own agentic tool loop (`StreamModelWithTools`); with `AgentHandler` bridging at the agent layer that loop is dead weight and a divergence source. Removed `Handler`, `StreamModel`, `StreamModelWithTools` + helpers, `ToolExecutor`, the static tool writers, and the `WithSystem/WithTools/WithMaxTurns` options. `handler.go` is now just the shared protocol primitives `AgentHandler` builds on; the model-level tests and the conformance suite (which existed only to exercise that handler) are gone, with the reusable `convertMessages` unit tests kept in `convert_test.go`.

Package builds, tests, lints, and license-checks clean; CI green. Agent-level conformance against the real `ai` package is a tracked follow-up.

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=1 · P2=2 · P3=0

The new agent adapter can pass client-forged system instructions into agent runs and can leave UI tool parts unresolved on terminal failures. It also deletes the conformance suite while leaving the task that invokes it behind.

Full review comments:

- [P1] Drop client-supplied system messages — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:83-83
  Because this reuses `convertMessages`, a browser request can include `role:"system"` in the posted history; `llmagent` only removes a leading system message before prepending the agent's prompt, and providers such as Google/Anthropic honor later system messages as system instructions. For any request with a non-leading system message (or a custom agent that doesn't strip them), the client can inject/override agent-level instructions despite the endpoint saying the agent supplies the system prompt. Filter or downgrade inbound system-role messages in `AgentHandler`.

- [P2] Close pending tool calls on terminal failures — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:366-366
  When a tool-calling run emits a tool request, this sends `tool-input-*` chunks but the adapter doesn't track the outstanding `toolCallId`. If the agent then terminates before a `ToolResponseEvent`—for example `ErrToolRegistry`, `FinishReasonInputRequired`, or any iterator error after the request—the terminal paths only emit `error`/`finish`, leaving `useChat` with a dynamic tool part stuck in `input-available` instead of transitioning to `output-error`. Track pending tool calls and close them before terminal error/finish chunks.

- [P2] Keep the conformance task runnable — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/conformance/run.sh:1-1
  Deleting this script removes the directory that `Taskfile.yaml`'s `test:conformance` target still uses as its working directory and runs via `bash run.sh`, so `task test:conformance` now fails immediately with a missing directory/script instead of running the AI SDK conformance suite. Restore/move the suite or update/remove the task target.

Script done on 2026-07-03 20:15:25+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=2 · P3=0

The adapter can emit invalid tool-output chunks for agents that rely on ToolRequestEvent, and custom tool error mapping receives JSON payloads instead of the actual error string. These are user-visible protocol/API issues that should be fixed before considering the patch correct.

Full review comments:

- [P2] Handle ToolRequestEvent before streaming tool outputs — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:349-350
  StreamAgent is exported for any `agent.Agent`, but this path drops `agent.ToolRequestEvent`. For agents that use the documented ToolRequestEvent to signal a tool call instead of also embedding the request in a `MessageEvent`, the subsequent `ToolResponseEvent` emits `tool-output-available` without a prior `tool-input-*` chunk, which violates UI Message Stream ordering and causes `useChat` to reject or strand the tool part. Emit/track these request events, while deduping LLMAgent's MessageEvent copy.

- [P2] Extract tool error text before calling onError — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:218-220
  When a registry-backed tool fails, `ToolResponsePart.Result` is JSON such as `{"error":"db connection failed"}`, not the raw `err.Error()`. With a custom `WithOnError` mapper that surfaces `err.Error()`, this sends that JSON object as `errorText`, regressing the documented tool-output-error mapper behavior; extract the error string from the payload before passing it to `onError`.

Script done on 2026-07-03 20:30:28+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=1 · P3=0

The adapter can emit duplicate stream error chunks for the normal agent error-event then error-finish sequence, causing incorrect client error handling.

Review comment:

- [P2] Avoid duplicating terminal error chunks — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:468-472
  When an agent follows the documented failure sequence (for example LLMAgent on content-filter/unknown finish: `ErrorEvent` followed by `InvocationEndEvent{FinishReasonError}`), the earlier `ErrorEvent` branch has already emitted an `error` chunk, and this branch emits a second generic one. `useChat`/`readUIMessageStream` will invoke `onError` twice and can mask the original mapped error; only synthesize this terminal error when no error chunk was already sent.

Script done on 2026-07-03 20:50:33+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=1 · P3=0

The adapter mishandles a supported ToolRequestEvent-only sequence by failing to close the tool-call step before the final answer, which corrupts the conversation history on the next turn.

Review comment:

- [P2] Close ToolRequestEvent-only steps before answers — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:346-348
  For agents that report a tool call only via `ToolRequestEvent` (which this path says it supports), this opens a step and leaves it open until the next `MessageEvent`. The resulting UI message puts the tool part and the post-tool answer in the same `step-start`; when the client resends that history, `reconstructAssistant` will serialize it as `assistant(tool_call, answer)` before `user(tool_result)`, reversing the tool-result/answer order for the next model request. Close the tool-call step or add a new step boundary before streaming post-tool content when no `MessageEvent` supplies it.

Script done on 2026-07-03 21:01:26+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=2 · P3=0

The new adapter can emit protocol-invalid tool output chunks for recovered tool calls and can report canceled invocations as normal finishes. These are runtime correctness issues for supported agent-streaming scenarios.

Full review comments:

- [P2] Guard tool responses without a matching input — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:328-330
  When `StreamAgent` is used with an invocation that triggers `llmagent`'s incomplete-tool-call recovery, the agent emits `ToolResponseEvent` before any `MessageEvent` for that call. This branch writes a `tool-output-*` chunk unconditionally, but the AI SDK parser requires a prior `tool-input-*` for the same `toolCallId`, so the client stream errors before the recovered answer can be processed. Since `pending` already tracks emitted inputs, skip or synthesize the input before writing the output.

- [P2] Return an abort for interrupted agent events — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:496-499
  If the request context is canceled before or between `LLMAgent` turns, `Run` yields `InvocationEndEvent{FinishReasonInterrupted}` with no iterator error, so this mapping sends a normal `finishReason: "other"` and `[DONE]` rather than the abort chunk used in the cancellation path. In server timeouts or pre-canceled requests that still have an open response, `useChat` will treat the canceled run as a completed assistant message; handle interrupted finishes with `ctx.Err()` as abort instead of a normal finish.

Script done on 2026-07-03 21:28:33+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=1 · P3=0

The new adapter drops valid tool request events for generic agent implementations, causing tool calls to disappear from the UI stream in that scenario.

Review comment:

- [P2] Handle ToolRequestEvent before tool outputs — /home/johannes/projects/ai-sdk-go/.claude/worktrees/vercel-adapter/adapter/vercelaisdk/uimessagestream/agent.go:372-378
  When `AgentHandler` is used with an `agent.Agent` implementation that reports tool calls via the documented `ToolRequestEvent` instead of also embedding them in a `MessageEvent` like `llmagent`, this branch drops the request, so `pending` is never populated and the subsequent `ToolResponseEvent` is skipped by `writeDynamicToolResponse`. The browser will never receive the tool-input/tool-output chunks for that call; either handle `ToolRequestEvent` with ID deduping or make the handler explicitly llmagent-only.

Script done on 2026-07-03 21:39:30+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## codex review loop — resolved, with one deliberate design decision

Ran codex iteratively; it drove the adapter from a P1 security issue down through progressively narrower P2 hardening on the **exported** `StreamAgent` under llmagent's unusual event orderings. Fixed and tested along the way: client system-message injection, incomplete/forged tool-call history, tool-error sanitization + payload unwrap, length→`length`, terminal-error dedup, orphan tool-output guard, interrupt→abort.

The loop then began **oscillating** on one genuine design tension: whether to handle `agent.ToolRequestEvent` for a hypothetical agent that reports tool calls *only* via that event and never in a `MessageEvent`. Handling it (codex pass 3) introduced step-ordering/dedup bugs (codex passes 4–5); reverting it (to match the sibling `a2a.Executor`, which sources tool calls from `MessageEvent`) brought codex back to "handle ToolRequestEvent" (pass 7).

**Decision:** `AgentHandler` follows the same MessageEvent-sourced contract as the shipped `adapter/a2a` `Executor`. Every `agent.Agent` in this SDK (llmagent) emits tool calls in `MessageEvent`, so all real usage is covered. This is now documented explicitly in `AgentHandler`'s doc comment as a deliberate, precedent-backed choice (codex's own suggested resolution: "make the handler explicitly llmagent-only"). Stopping the loop here — the remaining finding is a documented decision, not a defect.

Package builds, tests, lints, and license-checks clean; CI green.

adapter/vercelaisdk/uimessagestream/agent.go
```diff
@@ -0,0 +1,576 @@
...
+		}
+
+		// A cancellation between turns surfaces as InvocationEndEvent{interrupted}
+		// with no iterator error. Emit an abort (not a normal finish) so useChat
+		// treats it as aborted rather than a completed assistant message.
+		if end, ok := event.(agent.InvocationEndEvent); ok &&
+			(end.FinishReason == agent.FinishReasonInterrupted || ctx.Err() != nil) {
+			as.sw.writeAbort(ctx)
+			return
+		}
```
malinskibeniamin, line 432: **[P1 Major]** Cancellation/abort paths skip step + pending-tool close — dangling parts on `useChat`.

Both abort exits — the iterator-error-during-cancel path (`ctx.Err() != nil` → `writeAbort`, lines 286-289) and the `InvocationEndEvent{Interrupted}` / mid-loop-cancel path (lines 301-305) — jump straight to `as.sw.writeAbort(ctx)`, which only emits `abort` + `[DONE]` (handler.go:299-307). Neither calls `as.endStep()` nor `as.closePendingTools()`.

This directly violates the invariant documented on this very type (lines 128-129: "every terminal path closes the open step before the terminator"). Every OTHER terminal path — `terminate` (321) and `handleInvocationEnd` (464) — closes the step and resolves pending tools first; abort is the one that does not.

Why: if a client disconnects (or the agent is interrupted) after a `start-step`/`text-start`/`text-delta` was flushed, or after a `tool-input-available` for a call with no `ToolResponseEvent` yet, the browser is left with an unbalanced `start-step` (no `finish-step`) and/or a dynamic tool part stuck in `input-available`. That is the #1 dangling-part hazard — a hang/parse-desync in `useChat` — that the rest of this file is architected to prevent.

The existing test `TestAgentHandler_InterruptedFinishAborts` (agent_test.go:359) only exercises abort with NO open step/tool, so it passes even though these paths leak.

Suggested fix: before each `as.sw.writeAbort(ctx)`, close the open step and pending tools, e.g. add an `as.abort(ctx)` helper that does `_ = as.endStep(); as.closePendingTools()` then `as.sw.writeAbort(ctx)`, and call it at both sites. Add a test that opens a step (delta + pending tool-input) then cancels, asserting `finish-step` and `tool-output-error` precede `abort`.

One-shot prompt: In repo redpanda-data/ai-sdk-go, branch `jb/vercel-ui-message-stream-agent`, file adapter/vercelaisdk/uimessagestream/agent.go, add a method `func (as *agentStreamer) abort(ctx context.Context) { _ = as.endStep(); as.closePendingTools(); as.sw.writeAbort(ctx) }` and replace both `as.sw.writeAbort(ctx)` calls in StreamAgent (the ctx-cancel iterator-error branch and the InvocationEndEvent-interrupted branch) with `as.abort(ctx)`; add a test in agent_test.go that scripts an AssistantDeltaEvent + a MessageEvent carrying a ToolRequestPart, cancels the request context before InvocationEndEvent, and asserts the emitted chunk types include `finish-step` and `tool-output-error` before `abort`; verify with `go test ./adapter/vercelaisdk/uimessagestream/`.

adapter/vercelaisdk/uimessagestream/agent.go
```diff
@@ -0,0 +1,576 @@
...
+	as.terminate(onError(errors.New("incomplete agent run")))
+}
+
+// terminate closes the open step, resolves any tool call still pending (so the
+// client never strands a dynamic tool part in input-available), then emits the
+// error + finish{error} + [DONE] terminator.
+func (as *agentStreamer) terminate(errText string) {
+	_ = as.endStep()
+	as.closePendingTools()
+	_ = as.ew.WriteChunk(Chunk{"type": "error", "errorText": errText})
```
malinskibeniamin: **[P2 Minor]** `terminate` emits an `error` chunk unconditionally — can double-emit after a recoverable `ErrorEvent`.

`terminate` (line 324) always writes an `error` chunk, but `handleEvent` already sets `as.errored = true` and emits an `error` chunk on a recoverable `agent.ErrorEvent` (lines 373-374). If a recoverable `ErrorEvent` is followed by the stream ending WITHOUT an `InvocationEndEvent` (the incomplete-run path at line 315 → `terminate`), the client sees TWO `error` chunks and its `onError` fires twice — the same double-fire the `as.errored` guard in `handleInvocationEnd` (line 480) exists to prevent. Similarly, `handleInvocationEnd`'s `controlText != ""` branch (line 475) is not guarded by `as.errored`, so a recoverable `ErrorEvent` followed by `FinishReasonMaxTurns` also yields two error chunks.

Why: the error-terminal invariant is "at most one error chunk then finish{error}"; these two paths can emit error-then-error-then-finish.

Suggested fix: guard the `error` write in `terminate` with `if !as.errored`, and add the same `&& !as.errored` guard to the `controlText != ""` branch in `handleInvocationEnd` (still emitting `finish{error}` in both cases).

One-shot prompt: In repo redpanda-data/ai-sdk-go, branch `jb/vercel-ui-message-stream-agent`, file adapter/vercelaisdk/uimessagestream/agent.go, in `terminate` wrap the error-chunk write in `if !as.errored { ... }` (keep the finish{error} + WriteDone), and in `handleInvocationEnd` guard the `controlText != ""` branch so the error chunk is only written when `!as.errored`; add a test scripting an `agent.ErrorEvent` followed by no InvocationEndEvent asserting exactly one `error` chunk precedes `finish`; verify with `go test ./adapter/vercelaisdk/uimessagestream/`.

adapter/vercelaisdk/uimessagestream/handler.go
```diff
@@ -1,28 +1,41 @@
...
 // Known limitations:
-//   - Inbound file/image parts are not forwarded to the model: the llm.Part
-//     type has no file kind yet. Inbound reasoning parts are likewise dropped.
-//   - The handler calls model.GenerateEvents directly; interceptor plugins
-//     (retry, OTel) must be wired at the model level.
+//   - Inbound file/image parts are not forwarded: the llm.Part type has no file
+//     kind yet. Inbound reasoning parts are likewise dropped.
 //
 // Reference: https://github.com/vercel/ai
 package uimessagestream
```
malinskibeniamin, line 59: **[P1 Major]** Breaking removal of exported API (`Handler`, `StreamModel`, `StreamModelWithTools`, `ToolExecutor`, `WithSystem`, `WithTools`, `WithMaxTurns`) with no migration note.

This package doc was rewritten to describe only `AgentHandler`, but the commit "drop the model-level handler" removes seven exported symbols from a published module (`github.com/redpanda-data/ai-sdk-go`). Any external consumer that pinned `uimessagestream.Handler(model, ...)` / `StreamModelWithTools(...)` will fail to compile on upgrade with zero guidance. (In-repo this is self-consistent — the only caller was `conformance/testserver/main.go`, which this PR also deletes — so nothing in the repo breaks. The risk is purely external.)

**Why:** For a shared SDK, silently deleting exported symbols is a semver-major break. There is no CHANGELOG, deprecation shim, or migration note pointing model-level callers at `AgentHandler` (or telling them why the seam moved).

**Suggested fix:** Add a short migration note — either a `// Deprecated: ... use AgentHandler` doc-comment retained for one release, or a `Migration` paragraph in this package doc / a CHANGELOG entry — stating that the model-level `Handler`/`StreamModel*` were removed and external callers must move to `AgentHandler` (agent seam) and explaining the loop/telemetry reason from the PR description.

One-shot prompt: In repo redpanda-data/ai-sdk-go branch `jb/vercel-ui-message-stream-agent`, extend the package doc comment in adapter/vercelaisdk/uimessagestream/handler.go (lines 15-41) with a "Migration" paragraph noting that the model-level Handler/StreamModel/StreamModelWithTools/ToolExecutor/WithSystem/WithTools/WithMaxTurns exports were removed and external callers must switch to AgentHandler, and add a matching CHANGELOG entry.

Taskfile.yaml
```diff
@@ -73,12 +73,6 @@ tasks:
...
 
-  test:conformance:
-    desc: Run the Vercel AI SDK UI message stream conformance suite (requires node + bun)
-    dir: adapter/vercelaisdk/uimessagestream/conformance
-    cmds:
-      - bash run.sh
-
   security:
     desc: Run security vulnerability checks (govulncheck + osv-scanner)
     deps:
```
malinskibeniamin, line 84: **[P1 Major]** This PR deletes the entire real-SDK conformance suite instead of adding one.

Why: The PR description says this ships an `AgentHandler` "alongside the existing model-level `Handler`," implying the conformance harness is new/kept infrastructure guarding both. But diffing against `main` (PR base `8c666ae1`) shows the opposite: `main` already has `test:conformance` wired here (added in #132), and this PR's head branch (`d271ec3d`) removes the task entirely — confirmed via `gh api repos/redpanda-data/ai-sdk-go/pulls/164/files`, which reports `adapter/vercelaisdk/uimessagestream/conformance/{.gitignore,bun.lock,conformance.test.mjs,package.json,run.sh,testserver/main.go}` all as `removed`, plus the pre-existing `handler_test.go` (2445 lines) also `removed`. The model-level `uimessagestream.Handler` function these tests targeted is gone too — grep for `^func Handler` in `handler.go` at this SHA returns nothing. The replacement (`agent_test.go`, 434 lines) is Go-only: it hand-parses SSE with a local `parseSSEChunks` helper, not the real `@ai-sdk` client (`readUIMessageStream`/`DefaultChatTransport`), so there is no independent-of-the-implementation protocol validation left in the repo at all — the exact "fake success" failure mode this review was scoped to catch.

Suggested fix: restore `test:conformance` here and the conformance/ directory, then either (a) keep `Handler` and its tests and add a companion `/api/agent-*` conformance path exercising `AgentHandler` with a scripted fake agent (error/finish{error}, dynamic tool calls, step open/close, keep-alive ping), or (b) if `Handler` is intentionally being retired in favor of `AgentHandler`, retarget the existing conformance server/tests at `AgentHandler` rather than deleting them, so real-SDK SSE decoding coverage is not lost.

One-shot prompt: In redpanda-data/ai-sdk-go on branch `jb/vercel-ai-message-stream-agent`, restore `test:conformance` in `Taskfile.yaml` and the deleted `adapter/vercelaisdk/uimessagestream/conformance/` directory (git show main:Taskfile.yaml and main:adapter/vercelaisdk/uimessagestream/conformance/* to recover the content), then extend `conformance/testserver/main.go` and `conformance/conformance.test.mjs` with new endpoints/tests that wire `AgentHandler` (not just the retired `Handler`) against a scripted fake `agent.Agent` covering the error-terminal path, dynamic tool-call chunks, step open/close, and keep-alive ping, so real `@ai-sdk` client parsing continues to validate the SSE protocol.

adapter/vercelaisdk/uimessagestream/convert_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+// assertRolesAlternate verifies no two consecutive messages share a role, which
+// providers such as Anthropic require.
+func assertRolesAlternate(t *testing.T, msgs []llm.Message) {
+	t.Helper()
+
+	for i := 1; i < len(msgs); i++ {
+		assert.NotEqualf(t, msgs[i-1].Role, msgs[i].Role, "messages %d and %d must not share a role", i-1, i)
+	}
+}
```
malinskibeniamin: **[P1 Major]** Coverage regression: the deleted `handler_test.go` covered still-live primitives that were not re-homed here.

`convert_test.go` correctly inherited the five `TestConvertMessages_*` cases, but the deletion of `handler_test.go` (2445 lines) also dropped tests for code that is **still live** and still exercised by the new agent path:

- `TestWriteChunk_DoesNotHTMLEscape` — `EventWriter.WriteChunk` still calls `enc.SetEscapeHTML(false)` (handler.go:494), a deliberate byte-for-byte correctness property so deltas containing `<`, `>`, `&` (markdown/code) go over the wire unescaped. `agent.go` calls `WriteChunk` 14×. Nothing in `convert_test.go` or `agent_test.go` now asserts this — a silent regression here would corrupt every code-block response with `<`.
- `TestWithOnError_SurfacesCustomErrorText` — only the default-sanitize path is re-covered (`agent_test.go` `ToolErrorSanitized`); the custom-mapper-surfaces-text branch of `WithOnError` (handler.go:95) has no test.

**Why:** These functions did not move with the model handler; they are protocol primitives this file still owns and the agent handler depends on. Deleting their only tests drops live-path coverage.

**Suggested fix:** Re-home the two dropped primitive tests into `convert_test.go` (the new home for handler.go-primitive tests): a `TestWriteChunk_DoesNotHTMLEscape` asserting `<b>&"</b>` round-trips unescaped through `EventWriter.WriteChunk`, and a `TestWithOnError_SurfacesCustomErrorText` asserting a non-default `ErrorMapper` reaches the emitted error chunk.

One-shot prompt: In repo redpanda-data/ai-sdk-go branch `jb/vercel-ui-message-stream-agent`, add to adapter/vercelaisdk/uimessagestream/convert_test.go two `t.Parallel()` tests recovered from the deleted handler_test.go — `TestWriteChunk_DoesNotHTMLEscape` (an EventWriter over a buffer, WriteChunk a text-delta containing `<`, `>`, `&`, assert the raw SSE contains them unescaped) and `TestWithOnError_SurfacesCustomErrorText` (assert a custom ErrorMapper option is applied to config.onError).

adapter/vercelaisdk/uimessagestream/handler.go
birdayz: i will ignore this. the previous implementation was flawed

adapter/vercelaisdk/uimessagestream/agent.go
birdayz: Fixed in 428adfc1. Both abort exits now go through an abort() helper that closes the open step and resolves pending tool calls before the abort terminator, same cleanup as every other terminal path. Added a table test covering both triggers (iterator error after cancel, interrupted invocation end) with an open text span and a pending tool call, asserting text-end/finish-step/tool-output-error all precede abort.

adapter/vercelaisdk/uimessagestream/agent.go
birdayz: Fixed in 428adfc1. terminate() and the control-message branch in handleInvocationEnd now skip the error chunk when a recoverable ErrorEvent already emitted one, so the invariant is at most one error chunk then finish{error}. Test asserts exactly one error chunk for all three follow-ups: incomplete run, max-turns finish, error finish.

Taskfile.yaml
birdayz: Restored in 8c0012e6, retargeted at AgentHandler (option b). The test server now wires AgentHandler to llmagent over scripted fakellm models -- the full agent loop, not a protocol shim -- and test:conformance is back in the Taskfile. The suite keeps the old coverage and adds the agent-specific paths: dynamic-tool parts, agent-owned system prompt, mid-stream failure closing the open text part, keep-alive pings during tool exec, and a max-turns endpoint asserting exactly one terminal error with no tool part stuck in input-available. All 12 tests pass against the real ai@7.0.6 client (DefaultChatTransport + readUIMessageStream). Also verified end-to-end against a live OpenAI model through the same client.

adapter/vercelaisdk/uimessagestream/convert_test.go
birdayz: Re-homed. TestWriteChunk_DoesNotHTMLEscape is ported to convert_test.go in 1da4af5f; the custom-mapper branch of WithOnError is covered by TestWithOnError_SurfacesCustomErrorText in 428adfc1, rebuilt against AgentHandler (asserting the real error text reaches the emitted error chunk) since the model-level streamModelChunks harness is gone.

adapter/vercelaisdk/uimessagestream/handler.go
malinskibeniamin: Addressed in b58efec: the package documentation and README now include migration guidance from the removed model-level APIs to Handler(agent, store), including why the agent layer owns the loop.

malinskibeniamin:
## Revival update

Rebased onto current `main` and force-pushed with lease.

- migrated the adapter docs/examples to the server-authoritative session API
- rejected unknown UI roles and removed the dead conversion path
- updated migration guidance and conformance coverage
- verified unit/conformance tests, license headers, and golangci-lint

All GitHub checks are green; all review threads are resolved. Current head: `412ad9d9b871`.

PR #166, opened 2026-07-08
Add Amazon Nova 2 Lite to the Bedrock catalog

weeco:
Adds Amazon Nova 2 Lite so it surfaces in model discovery, pricing, and enforcement alongside the Claude entries.

- Registers the four inference-profile variants (global/us/eu/jp), inference-profile-only like the Claude models.
- Rates verified against the AWS Price List API — $0.30/$2.50 global, geo exactly 1.10×, cache-read 75% off, cache-write free.
- Relaxes the pricing test's cache-write assertion so free-cache-write models are allowed.

### DIFF
--- a/providers/bedrock/cross_region_test.go
+++ b/providers/bedrock/cross_region_test.go
@@ -59,6 +59,17 @@ func TestIsModelAllowedFromRegion(t *testing.T) {
 		{"au from ap-southeast-4 (Melbourne)", ModelClaudeHaiku45AU, "ap-southeast-4", true},
 		{"au from ap-southeast-6 (New Zealand)", ModelClaudeOpus46AU, "ap-southeast-6", true},
 
+		// Amazon Nova 2 Lite — same prefix-based routing as Claude (us/eu/jp
+		// geo profiles + global). Confirms the amazon. namespace routes like
+		// anthropic.
+		{"nova2 bare from us", ModelNova2Lite, "us-east-1", true},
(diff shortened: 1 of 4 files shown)


weeco:
@claude review

weeco:
Thanks — addressed in 9d3a6d0:

- 🔴 **Missing tests** — added \`TestLookupModel\` cases (global/us/eu/jp present; bare \`ModelNova2Lite\` and unpublished \`au.\` absent) and \`IsModelAllowedFromRegion\` cases for the Nova geo/global profiles.
- 🟡 **Pricing relaxation** — made it targeted: restored \`assert.Positive\` for cache-write on **all** models, with an explicit \`freeCacheWriteModels\` allow-list (the four Nova entries) that must be exactly zero. The forgot-\`WithCacheCreation\` guard is intact for Claude again.
- 🟡 **top_k** — dropped from \`SupportedParams\`; the Converse mapper only emits temperature/top_p/max_tokens/stop.
- 🟡 **Reasoning** — set to \`false\` with a comment: \`WithThinking\` emits the Anthropic-shaped thinking doc Nova rejects, so reasoning isn't reachable through the SDK yet (follow-up to wire Nova's schema).
- ℹ️ **Model-card values** — verified: 1M context / 64K output / image+video input from the Nova 2 Lite model card; input/output/cache-read from the AWS Price List bulk API (cache-write $0.00 → free).

malinskibeniamin (approved):

PR #167, opened 2026-07-09
bedrock: add Mistral Large 3 to the model catalog

malinskibeniamin:
Adds **Mistral Large 3** to the Bedrock model catalog, following the Nova 2 Lite pattern (ai-sdk-go#166). The model then surfaces in discovery, pricing, and enforcement, and the cloudv2 aigw can proxy it once the dep is bumped.

## What

- Registers the `us.mistral.mistral-large-3-675b-instruct` inference profile in `providers/bedrock/models.go` (ID constants + capability/constraint vars + catalog entry).
- Standard on-demand pricing: **$0.50/1M input, $1.50/1M output**, no prompt-cache pricing.
- Capabilities: streaming, tools, JSON mode + native structured output, system prompts, multi-turn. **No** vision (Mistral's multimodal entry is Pixtral) and **no** reasoning (that's Magistral). Context: 128K input / 8192 output.

## Why it doesn't fit the Claude/Nova mold

Two differences from every existing catalog entry, both handled deliberately:

1. **No geo/global split.** AWS prices Mistral Large 3 by *service tier* (standard/flex/priority/batch) under a single us-east-1 usagetype — there is no `global.` rate — so there's no global sibling and `TestGeoGlobalRatio` doesn't apply. Only the `us.` profile is registered; `eu.`/`global.` are deferred until confirmed by invocation (per the model-maintenance skill).
2. **No prompt caching.** The price list carries no cache-read/cache-write usagetype, so all cache rates stay zero — the documented shape for a non-caching provider (`pricing.Rates`). `TestAllModelsHavePricing` gains a `noCacheModels` set (parallel to `freeCacheWriteModels`) that asserts every cache rate is exactly zero for these models, while keeping Claude's "forgot `WithCacheCreation`" guard intact.

## Sources

- **AWS Price List bulk API** (authoritative; the pricing web page is JS-rendered and unusable): `offers/v1.0/aws/AmazonBedrock/20260707080509/us-east-1/index.json`, usagetype `mistral.mistral-large-3-675b-instruct-mantle-{input,output}-tokens-standard`.
- **LiteLLM** `bedrock_converse` catalog cross-check: input `5e-07`, output `1.5e-06` per token; 128K context; function calling + native structured output.

## Verification needed on an AWS account

This was authored without live Bedrock access, so a reviewer/tester with credentials should confirm:

- The exact invocation ID `us.mistral.mistral-large-3-675b-instruct` resolves (vs. a version-suffixed or bare form).
- Whether `eu.`/`global.` profiles exist (add entries if so).

The cloudv2 integration test (companion PR) skips cleanly when the model isn't accessible, matching the Nova 2 Lite test's behavior.

## Tests

- `go test ./providers/bedrock/...` — all green (271 tests).
- `golangci-lint run --new` — 0 issues. `gofmt`/`go vet` clean.
- Added lookup cases (us. profile present; bare + eu. absent) and the `noCacheModels` pricing branch.

---
🤖 Draft — part of a Bedrock model-introduction series. Companion cloudv2 PR bumps this dep and adds an `external_llm` integration test.

### DIFF
--- a/providers/bedrock/models.go
+++ b/providers/bedrock/models.go
@@ -137,25 +137,21 @@ const (
 
 // Model ID constants for Mistral AI models on Bedrock.
 //
-// Mistral Large 3 is served on Bedrock through a cross-region inference
-// profile, so — like the Claude and Nova entries — the bare
-// "mistral.mistral-large-3-675b-instruct" ID is registered only as a building
-// block for the prefixed variant and is not itself a catalog key (NewModel
-// always prepends the source region's geo prefix to an un-prefixed ID, so a
-// bare-only entry would be unreachable anyway).
-//
-// Only the us. profile is registered here. The AWS Price List publishes rates
-// solely under the us-east-1 usagetype
-// (mistral.mistral-large-3-675b-instruct-mantle-*); eu./global. profile
-// availability is not yet confirmed, and per the model-maintenance skill we do
-// not register a profile until it is verified by invocation. Add eu./global.
-// entries once confirmed. See
+// Unlike the Claude and Nova entries — which are inference-profile-only and
+// return ValidationException for the bare ID — Mistral Large 3 is an
+// on-demand / in-region model invoked by its BARE ID
+// "mistral.mistral-large-3-675b-instruct". It has no us./eu./global.
+// cross-region inference profile: the AWS Price List publishes rates only under
+// the single us-east-1 usagetype (mistral.mistral-large-3-675b-instruct-mantle-*)
+// with service tiers (standard/flex/priority/batch), and invoking the
+// "us.mistral..." profile returns "ValidationException: the provided model
+// identifier is invalid". So we register the bare ID and NewModel resolves it
+// as-is (no geo-prefix). See
 // https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards.html
 const (
-	// ModelMistralLarge3 is the bare Bedrock ID for Mistral Large 3
-	// (invoke via the prefixed variant).
-	ModelMistralLarge3   = "mistral.mistral-large-3-675b-instruct"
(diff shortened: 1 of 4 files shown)


malinskibeniamin:
@claude review

malinskibeniamin:
Thanks for the thorough review! This ran on the original commit; two follow-up commits address the findings:

**Callout 3 (invocation ID + `eu.`/`global.`)** — resolved in `90b104f`. The conformance integration test proved `us.mistral.mistral-large-3-675b-instruct` is invalid (`ValidationException: the provided model identifier is invalid`). Mistral Large 3 has **no cross-region inference profile** — the Price List publishes it only under a single us-east-1 service-tier usagetype, and LiteLLM lists only the bare id. So it's now registered as an **on-demand / in-region** model under the bare `mistral.mistral-large-3-675b-instruct`, and `NewModel` prefers an exact catalog match before auto-prefixing (backward-compatible; every existing cross-region profile still gets its geo prefix). The bare entry has no geo prefix, so it's correctly excluded from the per-region conformance sweep — **CI is now green**. The `eu.`/`global.` question is moot (no profiles exist).

**Callouts 1 & 2 (structured-output support, 8192 max-output)** — grounded in `34da6e7`. Both come from LiteLLM's `bedrock_converse` entry for `mistral.mistral-large-3-675b-instruct` (`supports_native_structured_output: true`; `max_output_tokens: 8192`). The comment now cites that source and flags that a live `Converse` call should still confirm constrained decoding + the exact output cap before this leaves draft — I authored without live Bedrock access, so those remain the reviewer-with-credentials checks.

**Minor observations** — left as-is per your note (the $0 cache-read rate is harmless for a non-caching model; the long catalog comment earns its length given the two departures from the Claude/Nova mold).

MikolajGagatekRP (approved):

PR #168, opened 2026-07-09
bedrock: add xAI Grok 4.3 to the model catalog

malinskibeniamin:
Adds **xAI Grok 4.3** to the Bedrock model catalog, following the Mistral Large 3 pattern (ai-sdk-go#167).

> **Note on "Grok 4.5":** it doesn't exist. xAI's version line runs 4 → 4.1 → 4.20 → **4.3**, and 4.3 is the latest — and the only Grok on Bedrock (`xai.grok-4.3`). This PR adds that.

## What

- Registers `xai.grok-4.3` as an **on-demand / in-region** model (bare ID, no inference profile) — same class as Mistral Large 3. AWS prices it by service tier under a single us-east-1 usagetype with no cross-region profile, and invoking a `us.xai…` profile returns `ValidationException`. `NewModel` prefers an exact catalog match before auto-prefixing, so the bare model is invoked as-is while every existing cross-region profile keeps its geo prefix.
- Pricing (standard on-demand): **$1.25/1M input, $2.50/1M output, $0.20/1M cache-read**. Cache **write is free** (no cache-write usagetype) → `freeCacheWriteModels`, same as Nova.
- Capabilities: streaming, tools, JSON mode + native structured output, **vision**, system prompts, multi-turn. Reasoning is `false` (the SDK's only reasoning lever, `WithThinking`, emits an Anthropic-shaped thinking doc non-Anthropic models reject). Context: 128K input / 16,384 output.

## Sources

- **AWS Price List bulk API**: `offers/v1.0/aws/AmazonBedrock/20260707080509/us-east-1/index.json`, usagetype `xai.grok-4.3-mantle-{input,output,cache-read}-tokens-standard`.
- **LiteLLM** `bedrock_mantle/xai.grok-4.3`: input `1.25e-06`, output `2.5e-06`, cache-read `2e-07`; 128K context; function calling + native structured output + vision.

## Verification needed on an AWS account

Authored without live Bedrock access (same as #167). A reviewer/tester with credentials should confirm the bare `xai.grok-4.3` invocation ID resolves and that structured output / vision behave over Converse. The bare entry has no geo prefix, so it's excluded from the per-region conformance sweep; the companion cloudv2 `external_llm` integration test skips cleanly when access isn't granted.

## Tests

- `go test ./providers/bedrock/...` — all green (271). `golangci-lint --new` — 0 issues. `gofmt`/`go build ./...` clean.
- Lookup cases (bare present, `us.` absent) + `TestNewModel_BareInRegionModelNotPrefixed`.

---
🤖 Draft — part of a Bedrock model-introduction series (companion to #167). A cloudv2 dep-bump + integration-test PR will follow once this merges.

### DIFF
--- a/providers/bedrock/models.go
+++ b/providers/bedrock/models.go
@@ -154,6 +154,30 @@ const (
 	ModelMistralLarge3 = "mistral.mistral-large-3-675b-instruct"
 )
 
+// Model ID constants for xAI (Grok) models on Bedrock.
+//
+// Like Mistral Large 3 (and unlike the inference-profile-only Claude/Nova
+// entries), Grok 4.3 is an on-demand / in-region model invoked by its BARE ID
+// "xai.grok-4.3": it has no us./eu./global. cross-region inference profile (the
+// AWS Price List publishes rates only under the single us-east-1 usagetype
+// xai.grok-4.3-mantle-* with service tiers standard/flex/priority/batch, and
+// invoking a "us.xai..." profile returns "ValidationException: the provided
+// model identifier is invalid"). So we register the bare ID and NewModel
+// resolves it as-is.
+//
+// Unlike Mistral, Grok 4.3 DOES bill prompt-cache reads with free cache writes —
+// the same pricing shape as Amazon Nova (so it belongs in freeCacheWriteModels,
+// not noCacheModels).
+//
+// Note: xAI's latest published model is 4.3 (the version line runs
+// 4 -> 4.1 -> 4.20 -> 4.3); there is no "4.5". See
+// https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards.html
+const (
+	// ModelGrok43 is the on-demand / in-region Bedrock ID for xAI Grok 4.3
+	// (invoked bare, with no inference-profile prefix).
+	ModelGrok43 = "xai.grok-4.3"
+)
+
 // ModelDefinition defines a model with its capabilities and constraints.
 type ModelDefinition struct {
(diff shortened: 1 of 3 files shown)


malinskibeniamin:
@claude review

MikolajGagatekRP (approved):

malinskibeniamin:
Thanks! Addressed and rebased onto `main` (now includes the merged #167):

- **Minor 1 — stale `freeCacheWriteModels` comments.** Fixed: the doc comment and the `TestAllModelsHavePricing` free-cache-write branch comment now name both Amazon Nova 2 Lite **and** xAI Grok 4.3.
- **Minor 2 — `Reasoning: false` is descriptive, not a guardrail.** Reworded the `grok43Caps` comment to say exactly that: `WithThinking` is not gated on the capability, so passing it to Grok would still emit the Anthropic-shaped thinking doc and be rejected — the flag documents "no reachable reasoning lever here" (matching Nova), it doesn't prevent the emission.
- **⚠️ Vision / StructuredOutput / JSONMode asserted from LiteLLM.** The comment now cites `bedrock_mantle/xai.grok-4.3` as the source and flags a live Converse round-trip (image block + `outputConfig.textFormat`) as the pre-un-draft gate. Same live-verification caveat as #167 — I authored without AWS access.

Also renamed the new test to `TestNewModel_Grok43BareNotPrefixed` so it doesn't collide with the `TestNewModel_BareInRegionModelNotPrefixed` that landed with #167. `go test ./providers/bedrock/...` green (276), `golangci-lint --new` 0 issues.

PR #170, opened 2026-07-10
feat(bedrock): add Google Gemma 4 models on the mantle endpoint

weeco:
Adds the Google Gemma 4 family (31B, 26B-A4B and E2B) to the Bedrock catalog. Because these models only run on Bedrock's mantle endpoint over the OpenAI Responses API, the provider now routes them through a small SigV4 transport signed with the bedrock-mantle service name that reuses our existing OpenAI model instead of Converse. It's unit tested and I also verified it end to end against a real Gemma 4 model in us-east-2.

### DIFF
--- a/providers/bedrock/mantle.go
+++ b/providers/bedrock/mantle.go
@@ -1,0 +1,243 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


weeco:
@claude Review

weeco:
On the EU regional pricing suggestion: keeping Gemma at US standard-tier only, on purpose for now. A region override would actually be inert today: the cloudv2 spending resolver builds the pricing selector with Speed only, and nothing populates InferenceRegion on the Bedrock/mantle path (only the Anthropic mapper sets it), so a WithOverride(Selector{Region:...}) would compile but silently never match. Region-accurate Bedrock pricing is better done once in the cloudv2 spending pipeline (thread the served region into the selector), where it also fixes Nova and the Claude geo profiles, not just Gemma. Tracking that as a separate follow-up rather than adding an override that reads as handled while under-costing EU exactly as before.

simon0191 (approved):

PR #174, opened 2026-07-13
feat(models): add lifecycle and upgrade metadata

malinskibeniamin:
## Summary

Adds factual, provider-scoped catalog metadata so consumers can group model families, label confirmed lifecycle state, and derive the newest useful model without maintaining mutable frontier, modern, or legacy tags.

Older callable identifiers remain supported. Catalog-only records classify retired and end-of-life snapshots without advertising them for new selections.

## Catalog contract

**ModelCatalogMetadata** provides:

- **FamilyKey** — stable product family for UI grouping
- **UpgradeGroup** — provider-scoped track whose releases can be compared like-for-like; Bedrock routing geography is part of the track
- **ReleaseDate** — typed UTC-midnight time.Time availability date used for ordering
- **EndOfLifeDate** — optional typed UTC-midnight time.Time retirement schedule
- **Lifecycle** — one zero-safe UNKNOWN / ACTIVE / DEPRECATED / RETIRED enum
- **ProviderReplacement** — provider-named migration hint, not a latest-model signal
- **OfficialSourceURL** — provider evidence for the record
- **MetadataVerifiedDate** — typed UTC-midnight time.Time maintainer audit date with no lifecycle meaning

**ModelCatalogProvider** resolves discovered canonical identifiers, aliases, and dated snapshots. **ModelCatalogOverridesProvider** enumerates exact compatibility records that remain resolvable but are intentionally absent from Models().

## Recommendation and lifecycle semantics

Consumers derive the current choice as follows:

- compare non-zero ReleaseDate values within the same provider-scoped UpgradeGroup
- recommend only models whose Lifecycle is explicitly ACTIVE
- treat equal active release dates as tied and current
- keep older callable models available for compatibility
- change Lifecycle only from explicit provider evidence
- never infer lifecycle state from EndOfLifeDate alone, even after the date passes
- do not use ProviderReplacement for ranking
- fail open when grouping or date metadata is missing

Preview status remains represented by provider model identifiers; this PR does not add a separate release-stage enum.

## Provider audit and model corrections

The catalog covers OpenAI, Anthropic, Google Gemini, and Amazon Bedrock using official provider documentation.

Notable updates include:

- current OpenAI GPT flagship, balanced, efficient, and pro upgrade tracks
- gpt-5.3-codex preserved as a callable historical identifier but folded into the GPT flagship upgrade track
- GPT-5.5 Pro capabilities, constraints, pricing, and non-streaming behavior
- Gemini 3.1 Flash-Lite and current Gemini recommendation tracks
- published Bedrock regional profiles for Claude Opus, Sonnet, and Haiku
- exact retired and end-of-life OpenAI and Gemini compatibility records
- strict Anthropic catalog IDs without narrowing runtime custom-model compatibility
- documented OpenAI US/EU regional pricing
- typed catalog dates and one lifecycle enum instead of raw strings and two booleans

## Runtime correctness fixes included

- LLMAgent dispatches from ModelCapabilities.Streaming, so non-streaming models use Generate rather than GenerateEvents.
- OpenAI providers select US/EU inference-region pricing explicitly or derive it from the regional API hostname.
- OpenAI responses propagate the selected inference region for downstream pricing.

## Official sources

- [OpenAI models](https://developers.openai.com/api/docs/models/all) and [deprecations](https://developers.openai.com/api/docs/deprecations)
- [Anthropic models](https://platform.claude.com/docs/en/about-claude/models/overview), [model ID rules](https://platform.claude.com/docs/en/about-claude/models/model-ids-and-versions), and [deprecations](https://platform.claude.com/docs/en/about-claude/model-deprecations)
- [Google Gemini models](https://ai.google.dev/gemini-api/docs/models), [deprecations](https://ai.google.dev/gemini-api/docs/deprecations), and [release notes](https://ai.google.dev/gemini-api/docs/changelog)
- [Amazon Bedrock model lifecycle](https://docs.aws.amazon.com/bedrock/latest/userguide/model-lifecycle.html)

## Verification

- go test ./... — passed locally
- task license:check — passed locally
- task lint:new — 0 issues
- Check License Headers, Golangci Lint, and Test passed on 5d7c0d088866

## Consumer

[redpanda-data/cloudv2#28132](https://github.com/redpanda-data/cloudv2/pull/28132) derives its public recommendation contract from UpgradeGroup, typed ReleaseDate, and ModelLifecycle. It pins v0.0.0-20260714230211-5d7c0d088866. Audit provenance and provider migration hints remain internal to this SDK.


malinskibeniamin:
## Final audit and verification

- Adds typed, additive catalog metadata for OpenAI, Anthropic, Gemini, and Bedrock: family, recommendation tier, lifecycle, release stage, retirement/EOL data, replacement, source, and verification date.
- Marks a model retired only when the provider confirms retirement; an elapsed announced date alone is not enough.
- Resolves aliases and snapshots back to provider catalog metadata without guessing unsupported IDs.
- Updates defaults and docs toward current frontier/useful family leaders while preserving old identifiers for compatibility.
- Includes the current GPT-5.5 Pro capability/pricing definition and rejects unsupported streaming before an HTTP request.
- Consumer implementation: redpanda-data/cloudv2#28109.

### Verification

- Full Go test suite: green
- License headers: green
- golangci-lint: green
- PR CI: green

providers/openai/catalog.go
```diff
@@ -0,0 +1,99 @@
...
+const (
+	openAIModelsSource       = "https://developers.openai.com/api/docs/models"
+	openAIDeprecationsSource = "https://developers.openai.com/api/docs/deprecations"
+	openAIVerifiedDate       = "2026-07-13"
+)
+
+// ModelCatalog returns recommendation and lifecycle metadata for a canonical,
+// aliased, or dated OpenAI model identifier.
+func (*Provider) ModelCatalog(model string) (llm.ModelCatalogMetadata, bool) {
+	return openAIModelCatalog(resolveModelFamily(model))
```
malinskibeniamin: [P1 Major] Preserve exact OpenAI snapshot lifecycle

What: ModelCatalog collapses every supported prefix before the catalog lookup, so exact IDs lose their own lifecycle. For example, gpt-5-2025-08-07 and o3-2025-04-16 inherit the base aliases, while already-shut-down siblings such as gpt-4o-realtime-preview inherit gpt-4o metadata and the wrong replacement.

Why: OpenAI deprecates specific snapshots independently (https://developers.openai.com/api/docs/deprecations). This can mark a current alias deprecated, fail to mark a retired snapshot retired, or send users to the wrong replacement—breaking the catalog's confirmed-status contract and the UI's legacy filtering.

Suggested fix: Resolve explicit aliases and exact published snapshots first; only accept a strict, validated dated-snapshot suffix for family fallback, and return false for unrelated/custom siblings. Add positive and negative public ModelCatalog tests.

One-shot prompt: In ai-sdk-go on ben-malinski/model-catalog/lifecycle-metadata, update the OpenAI catalog/resolver to preserve exact alias and snapshot lifecycle metadata with strict suffix validation and regression tests, then run task test:unit.

providers/google/catalog.go
```diff
@@ -0,0 +1,74 @@
...
+const (
+	googleModelsSource       = "https://ai.google.dev/gemini-api/docs/models"
+	googleDeprecationsSource = "https://ai.google.dev/gemini-api/docs/deprecations"
+	googleVerifiedDate       = "2026-07-13"
+)
+
+// ModelCatalog returns recommendation and lifecycle metadata for a canonical,
+// aliased, or versioned Gemini model identifier.
+func (*Provider) ModelCatalog(model string) (llm.ModelCatalogMetadata, bool) {
+	return googleModelCatalog(resolveModelFamily(model))
```
malinskibeniamin: [P1 Major] Do not collapse retired Gemini previews into stable models

What: Prefix resolution maps gemini-3.1-flash-lite-preview to gemini-3.1-flash-lite, reporting it as stable/deprecated through 2027 even though the preview shut down on May 25, 2026. The same happens to retired gemini-2.5-pro-preview-* IDs.

Why: Google's official schedule distinguishes stable IDs from independently retired previews (https://ai.google.dev/gemini-api/docs/deprecations). Existing resources using a dead preview will therefore receive false lifecycle/release-stage metadata, defeating the confirmed-retirement rule and UI warnings.

Suggested fix: Match exact retired preview IDs before family fallback, explicitly map supported aliases, validate only documented version suffixes, and add retired-preview plus unknown-sibling tests.

One-shot prompt: In ai-sdk-go on ben-malinski/model-catalog/lifecycle-metadata, update the Google catalog/resolver to preserve exact preview lifecycle and release-stage metadata while rejecting unknown siblings, add regression tests, then run task test:unit.

providers/openai/model.go
```diff
@@ -86,6 +86,12 @@ func (m *Model) Generate(ctx context.Context, req *llm.Request) (*llm.Response,
 // It returns a Go 1.23+ iterator for streaming LLM responses.
 func (m *Model) GenerateEvents(ctx context.Context, req *llm.Request) iter.Seq2[llm.Event, error] {
 	return func(yield func(llm.Event, error) bool) {
+		if !m.definition.Capabilities.Streaming {
```
malinskibeniamin, line 89: [P1 Major] Fall back to non-streaming generation for GPT-5.5 Pro

What: GPT-5.5 Pro correctly declares Streaming: false, but every llm.Model implements EventsGenerator; LLMAgent.generate therefore always calls this method, which now returns ErrUnsupportedFeature, and never reaches Generate.

Why: The newly recommended Pro model is unusable through the normal agent path even though its non-streaming API is supported. This fails before any OpenAI request is made.

Suggested fix: Make agent dispatch check model.Capabilities().Streaming before choosing GenerateEvents, falling back to Generate when false, and cover GPT-5.5 Pro with an LLMAgent regression test.

One-shot prompt: In ai-sdk-go on ben-malinski/model-catalog/lifecycle-metadata, make LLMAgent select streaming from Capabilities().Streaming, verify GPT-5.5 Pro uses Generate, add a regression test, then run task test:unit.

providers/openai/models.go
```diff
@@ -393,6 +393,27 @@ var supportedModels = map[string]ModelDefinition{
...
+			Reasoning:        true,
+		},
+		Constraints: llm.ModelConstraints{
+			MaxInputTokens:    1_050_000,
+			MaxOutputTokens:   128_000,
+			SupportedParams:   []string{"max_tokens", "reasoning_effort", "reasoning_summary"},
+			MutuallyExclusive: [][]string{},
+		},
+		SupportedReasoningEfforts: []ReasoningEffort{ReasoningEffortMedium, ReasoningEffortHigh, ReasoningEffortXHigh},
+		Pricing:                   pricing.FlatInfo(30.00, 180.00, 0),
```
malinskibeniamin: [P1 Major] Account for GPT-5.5 Pro regional pricing

What: This flat rate has no region override, so pricing resolves to $30/M input and $180/M output even when the calculation selects a regional/data-residency endpoint.

Why: OpenAI documents a 10% uplift for GPT-5.5 Pro regional processing (https://developers.openai.com/api/docs/models/gpt-5.5-pro). The SDK currently underreports those requests by 10%, and the repository instructions explicitly require selector overrides when pricing varies by region.

Suggested fix: Add the 10%-uplift pricing.WithOverride entries for the canonical regional selector values, ensure OpenAI response/provider metadata supplies the matching region, and test both default and regional totals.

One-shot prompt: In ai-sdk-go on ben-malinski/model-catalog/lifecycle-metadata, wire GPT-5.5 Pro's documented 10% regional uplift through canonical pricing region selectors and OpenAI metadata, add default/regional cost tests, then run task test:unit.

providers/bedrock/models.go
```diff
@@ -701,6 +723,15 @@ var supportedModels = map[string]ModelDefinition{
 			pricing.NewRates(3.30, 16.50, 0.33).WithCacheCreation(4.125, 6.60, 0),
 		),
 	},
+	ModelClaudeSonnet46JP: {
+		Name:         ModelClaudeSonnet46JP,
+		Label:        "Claude Sonnet 4.6 (JP)",
+		Capabilities: claudeStandardCaps,
+		Constraints:  claudeContext200kConstraints,
```
malinskibeniamin: [P1 Major] Use Sonnet 4.6's published 1M context limit

What: The new JP Sonnet 4.6 profile uses claudeContext200kConstraints, limiting input to 200K tokens.

Why: AWS publishes a 1M-token context window and 64K max output for this exact profile (https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-4-6.html). SDK validation will reject otherwise valid 200K–1M requests before they reach Bedrock.

Suggested fix: Define/use Sonnet 4.6-specific 1M-input/64K-output constraints for JP and the other Sonnet 4.6 profiles, with discovery/validation tests.

One-shot prompt: In ai-sdk-go on ben-malinski/model-catalog/lifecycle-metadata, apply 1M-input/64K-output constraints to all Bedrock Sonnet 4.6 profiles including JP, add boundary tests, then run task test:unit.

providers/google/models.go
```diff
@@ -118,6 +119,28 @@ var supportedModels = map[string]ModelDefinition{
...
+			Reasoning:        true,
+		},
+		Constraints: llm.ModelConstraints{
+			TemperatureRange:  [2]float64{0.0, 2.0},
+			MaxInputTokens:    1048576,
+			MaxOutputTokens:   65536,
+			SupportedParams:   []string{"temperature", "top_p", "top_k", "max_tokens", "stop", "presence_penalty", "frequency_penalty"},
+			MutuallyExclusive: [][]string{},
+		},
+		Pricing: pricing.FlatInfo(0.25, 1.50, 0.025),
```
malinskibeniamin, line 144: [P3 Patch] Add required dollar annotations to new pricing tuples

What: New model prices are added as positional numeric arguments without the dollar-per-million comments required by CLAUDE.md; this line and the new OpenAI and Bedrock tuples have the same omission.

Why: These dense input/output/cache and cache-write positions are easy to transpose during future audits. The repository standard explicitly requires a dollar comment on every value.

Suggested fix: Annotate every newly added pricing argument with its $…/M meaning, using multiline calls where needed; no behavior change.

One-shot prompt: In ai-sdk-go on ben-malinski/model-catalog/lifecycle-metadata, add the required dollar-per-million annotations to every new Google, OpenAI, and Bedrock pricing value, then run task lint:new and task test:unit.

providers/openai/catalog.go
malinskibeniamin: Fixed in f4b0dd4. ModelCatalog now checks exact snapshot metadata first, resolves explicit aliases, only accepts strict YYYY-MM-DD snapshot suffixes, and rejects unrelated siblings. Added positive/negative public catalog regression tests.

providers/google/catalog.go
malinskibeniamin: Fixed in f4b0dd4. Gemini now handles exact retired previews before family fallback, maps gemini-flash-latest to gemini-3.5-flash, marks confirmed retired IDs explicitly, keeps stable 3.1 Flash Lite active, accepts only documented numeric version shapes, and rejects unknown siblings. Added regression tests.

providers/openai/model.go
malinskibeniamin: Fixed in b5ab384. LLMAgent dispatch now uses Capabilities().Streaming as the authoritative signal and falls back to Generate for non-streaming models. Added a regression model whose GenerateEvents fails to prove the batch path is selected.

providers/openai/models.go
malinskibeniamin: Fixed in 3312d8f and c39d059. GPT-5.5 Pro now has 10% US/EU pricing overrides, tested against global and regional totals. OpenAI regional endpoints are inferred from us.api.openai.com/eu.api.openai.com; WithInferenceRegion supports gateways, and responses propagate InferenceRegion for pricing selection.

providers/bedrock/models.go
malinskibeniamin: Fixed in 3312d8f. All five Bedrock Sonnet 4.6 profiles now share exact 1M-input/64K-output constraints. Added a regression test covering global, US, EU, AU, and JP discovery definitions.

providers/google/models.go
malinskibeniamin: Fixed in 3312d8f. Added dollar-per-million annotations for every newly introduced Google, OpenAI, and Bedrock pricing tuple, including cache read/write positions and regional rates. task lint:new reports 0 issues.

malinskibeniamin:
@claude review

malinskibeniamin:
All review threads resolved. CI is green.

malinskibeniamin:
Final review/fix pass complete.

- Replaced mutable `positioning` and release-stage labels with catalog facts: `family_key`, `recommendation_group`, required ISO `release_date`, optional EOL/replacement/source/verification metadata, and confirmed `retired` status.
- Recommendation order is now derived from release dates within each provider-scoped family group; ties remain visible and malformed metadata fails open.
- Added catalog validation for dates, retirement evidence, replacements, and official sources.
- All review threads resolved; no outstanding change requests.

CI is green: license headers, GolangCI-Lint, and tests.

malinskibeniamin:
Follow-up applied: `gpt-5.3-codex` is no longer modeled as a separate current Coding family.

- preserved the exact model ID for backward compatibility
- moved it to the normal `gpt` / `openai-gpt-flagship` upgrade track
- set `gpt-5.6-sol` as its replacement
- removed it from the derived recommendation set

The recommended OpenAI tracks are now GPT-5.6 Sol, Terra, Luna, plus GPT-5.5 Pro. SDK CI is green.

malinskibeniamin:
Lifecycle metadata follow-up is complete:

- added explicit, provider-confirmed `Deprecated` independent from `EndOfLifeDate` and `Retired`
- enforced `Retired => Deprecated` across the audited catalog
- audited OpenAI, Anthropic, Google Gemini, and Bedrock facts against official sources
- folded discontinued Codex recommendations into GPT IDs without the `-codex` suffix
- added regressions for deprecated-without-EOL, scheduled-EOL-without-inference, and exact retired overrides
- resolved all review threads; fresh-eyes and simplification passes found no remaining actionable issue

Verification: `go test ./...`; `task lint:new` (0 issues). GitHub license, lint, and test checks are all green at `6b28fc11187d`.

malinskibeniamin:
Final implementation and verification update:

- Model metadata now separates stable family identity from upgrade-track ranking.
- ReleaseDate drives latest-model selection; EndOfLifeDate remains informational.
- Deprecated and Retired are confirmed provider facts, so a past date alone never retires a model.
- ProviderReplacement is only a provider migration hint, not a ranking signal.
- OfficialSourceURL and MetadataVerifiedDate keep lifecycle assertions auditable.
- The discontinued Codex line is folded into the GPT upgrade track; no current model uses a -codex suffix.

Verification:
- go test ./... passed.
- task lint:new passed with 0 issues.
- All PR checks are green.
- Unresolved review threads: 0.

Cloud consumer: redpanda-data/cloudv2#28132.

providers/anthropic/catalog.go
```diff
@@ -0,0 +1,76 @@
...
+const (
+	anthropicModelsSource         = "https://platform.claude.com/docs/en/about-claude/models/overview"
+	anthropicDeprecationsSource   = "https://platform.claude.com/docs/en/about-claude/model-deprecations"
+	anthropicMetadataVerifiedDate = "2026-07-14"
+)
+
+// ModelCatalog returns factual catalog metadata for a canonical,
+// aliased, or dated Anthropic model identifier.
+func (*Provider) ModelCatalog(model string) (llm.ModelCatalogMetadata, bool) {
+	return anthropicModelCatalog(resolveModelFamily(model))
```
malinskibeniamin: [P1 Major] Reject unrecognized Anthropic siblings

What: `ModelCatalog` delegates to the runtime prefix resolver, so any suffix is accepted. I reproduced `claude-opus-4-8-custom` returning active Opus 4.8 metadata. Anthropic documents 4.6+ IDs as exact dateless pinned IDs; only pre-4.6 IDs use `YYYYMMDD` snapshots ([model ID rules](https://platform.claude.com/docs/en/about-claude/models/model-ids-and-versions)).

Why: This violates the PR's fail-open rule for unknown/invalid IDs. Catalog consumers can label a typo or custom ID active/current and resolve it to the canonical model.

Suggested fix: Give catalog lookup a strict resolver: exact IDs for 4.6+, and only explicit aliases or validated `YYYYMMDD` snapshots for older families; return false for other siblings. Add positive and negative public catalog tests.

One-shot prompt: In redpanda-data/ai-sdk-go on ben-malinski/model-catalog/lifecycle-metadata, make `providers/anthropic/catalog.go:26-29` reject unrecognized model siblings while preserving documented aliases/snapshots, add regression tests, then run `task test:unit`.

providers/bedrock/catalog.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func bedrockModelCatalog(name string) (llm.ModelCatalogMetadata, bool) {
+	route := bedrockRoutingKey(name)
+
+	switch name {
+	case ModelClaudeFable5Global, ModelClaudeFable5US:
+		return bedrockCatalog("claude-fable", "bedrock-claude-fable-"+route, "2026-06-09", "", bedrockFable5Source), true
+	case ModelClaudeFable5EU:
+		return bedrockCatalog("claude-fable", "bedrock-claude-fable-"+route, "2026-06-09", "", bedrockFable5Source), true
```
malinskibeniamin, line 57: [P1 Major] Stop recommending the unavailable Fable EU profile

What: This returns active metadata for `eu.anthropic.claude-fable-5`, and `Models()` still advertises it. I reproduced `discovered=true`, `Deprecated=false`, and `Retired=false`; because it is the only member of its EU upgrade group, release-date consumers mark it current. AWS publishes only `us.` and `global.` inference IDs for Fable 5 ([model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-fable-5.html#model-card-anthropic-claude-fable-5-programmatic-access)).

Why: EU users can be offered/recommended an ID AWS does not accept. The nearby comment says the entry is hidden from new selections, but the catalog currently does the opposite.

Suggested fix: Separate runtime-compatibility lookup from discovery so the EU ID is omitted from `Models()` and cannot become recommended, without inventing a deprecation claim. Cover discovery/recommendation and retained runtime compatibility.

One-shot prompt: In redpanda-data/ai-sdk-go on ben-malinski/model-catalog/lifecycle-metadata, stop `ModelClaudeFable5EU` from being advertised or recommended while preserving any required runtime compatibility, add regression tests, then run `task test:unit`.

providers/openai/catalog.go
```diff
@@ -0,0 +1,181 @@
...
+
+	if alias, ok := modelAliases[model]; ok {
+		return openAIModelCatalog(alias)
+	}
+
+	if _, ok := supportedModels[model]; ok {
+		return openAIModelCatalog(model)
+	}
+
+	if family, ok := openAIDatedSnapshotFamily(model); ok {
```
malinskibeniamin, line 61: [P1 Major] Preserve GPT-4o snapshot shutdown facts

What: The exact override map omits `gpt-4o-2024-05-13`, so this fallback returns base `gpt-4o` metadata. I reproduced `EndOfLifeDate=""` and `ProviderReplacement="gpt-5.6-sol"`; OpenAI lists this snapshot for shutdown on 2026-10-23 with `gpt-5.5` as the substitute ([deprecation table](https://developers.openai.com/api/docs/deprecations#2026-04-22-legacy-gpt-model-snapshots)).

Why: Existing resources lose the provider-confirmed retirement schedule and exact migration hint, breaking the catalog's snapshot-lifecycle contract.

Suggested fix: Add an exact override before family fallback (and audit other officially scheduled snapshots whose facts differ from their base alias), then cover the exact ID and override enumeration.

One-shot prompt: In redpanda-data/ai-sdk-go on ben-malinski/model-catalog/lifecycle-metadata, add exact OpenAI lifecycle metadata for `gpt-4o-2024-05-13` and other audited snapshots that differ from base aliases, add regression tests, then run `task test:unit`.

malinskibeniamin (commented):
Fresh review of the current head; findings are inline.

providers/google/catalog.go
```diff
@@ -0,0 +1,136 @@
...
+
+const (
+	googleModelsSource         = "https://ai.google.dev/gemini-api/docs/models"
+	googleDeprecationsSource   = "https://ai.google.dev/gemini-api/docs/deprecations"
+	googleMetadataVerifiedDate = "2026-07-14"
+)
+
+var googleExactModelCatalogOverrides = map[string]llm.ModelCatalogMetadata{
+	"gemini-3.1-flash-lite-preview": googleCatalog("gemini-flash-lite", "2026-03-03", "2026-05-25", true, true, ModelGemini31FlashLite, googleDeprecationsSource),
+	"gemini-2.0-flash":              googleCatalog("gemini-flash", "2025-02-05", "2026-06-01", true, true, ModelGemini35Flash, googleDeprecationsSource),
```
malinskibeniamin: [P1 Major] Enumerate the retired Gemini 2.0 exact snapshot

What: The override map contains `gemini-2.0-flash` but not the official exact ID `gemini-2.0-flash-001`. I reproduced `ModelCatalog("models/gemini-2.0-flash-001")` returning false. Google lists both IDs as shut down on 2026-06-01 with `gemini-3.5-flash` as replacement ([deprecation table](https://ai.google.dev/gemini-api/docs/deprecations#gemini-20-models)).

Why: Existing resources using the pinned stable ID receive no retired status or replacement, which defeats the catalog-only compatibility path added by this PR.

Suggested fix: Enumerate `-001` and audit the other exact retired IDs in the covered Gemini families; add public lookup and override-enumeration tests.

One-shot prompt: In redpanda-data/ai-sdk-go on ben-malinski/model-catalog/lifecycle-metadata, add exact retired Gemini lifecycle overrides including `gemini-2.0-flash-001`, audit covered-family exact IDs, add regression tests, then run `task test:unit`.

providers/google/catalog.go
```diff
@@ -0,0 +1,136 @@
...
+	"gemini-2.5-pro-preview-03-25":  googleCatalog("gemini-pro", "2025-03-25", "2025-12-02", true, true, ModelGemini31ProPreview, googleDeprecationsSource),
```
malinskibeniamin: [P2 Minor] Reconcile the 03-25 preview's release date with its evidence

What: This record stores `2025-03-25`, but its cited deprecations page lists March 3, while Google's release notes say the exact public-preview ID launched April 4 (March 25 was the `-exp-03-25` model) ([deprecations](https://ai.google.dev/gemini-api/docs/deprecations#gemini-25-pro-models), [release notes](https://ai.google.dev/gemini-api/docs/changelog#april-4-2025)).

Why: `ReleaseDate` is documented as provider availability and `OfficialSourceURL` as its audit evidence; this value is supported by neither source, so consumers receive a factual date with broken provenance.

Suggested fix: Define whether this field tracks exact-ID availability or underlying model lineage, then use the matching official date/source and lock it with a test.

One-shot prompt: In redpanda-data/ai-sdk-go on ben-malinski/model-catalog/lifecycle-metadata, reconcile `gemini-2.5-pro-preview-03-25` release-date semantics with official Google evidence in `providers/google/catalog.go:34`, update the source/value and test, then run `task test:unit`.

providers/anthropic/catalog.go
malinskibeniamin: Fixed in ec5fcf8. Catalog lookup now uses a strict Anthropic-only resolver: exact supported IDs are accepted, while only the documented pre-4.6 families accept validated YYYYMMDD snapshots. The permissive runtime resolver remains unchanged for compatibility. Added positive/negative regression coverage, including the custom sibling repro.

providers/bedrock/catalog.go
malinskibeniamin: Fixed in ec5fcf8. Provider.Models now omits eu.anthropic.claude-fable-5 so it cannot be offered or recommended, while lookupModel still retains the ID for existing runtime configurations. Discovery and compatibility regressions cover both sides.

providers/openai/catalog.go
malinskibeniamin: Fixed in ec5fcf8. Added the exact gpt-4o-2024-05-13 shutdown record (2026-10-23, replacement gpt-5.5) and audited/added the other covered OpenAI snapshots whose lifecycle facts differ from their base aliases: GPT-3.5 Turbo, GPT-4 Turbo, O1 Pro, and O4 Mini. Exact lookup and override-enumeration tests lock the records.

providers/google/catalog.go
malinskibeniamin: Fixed in ec5fcf8. Added gemini-2.0-flash-001 plus the other covered exact retired Flash and Flash-Lite IDs from Google’s deprecation tables. Public lookup (including models/ prefix), lifecycle/replacement facts, and override enumeration are covered by regressions.

providers/google/catalog.go
malinskibeniamin: Fixed in ec5fcf8. The exact public-preview ID now records 2025-04-04, with Google’s April 4 changelog entry as OfficialSourceURL. The code distinguishes it from the separate -exp-03-25 identifier, and tests lock both the date and evidence URL.

malinskibeniamin:
Review follow-up complete in `ec5fcf84eaa9`.

- Resolved all 5 inline findings: strict Anthropic catalog IDs, non-discoverable Fable EU compatibility entry, exact OpenAI snapshot lifecycle facts, exact retired Gemini IDs, and corrected Gemini 2.5 Pro March preview evidence.
- Addressed the remaining review observations: all covered OpenAI models with documented regional pricing now have US/EU overrides; `NewResponseMapper` now takes one explicit region; unknown current Gemini preview snapshots remain intentionally unclassified and have a negative regression.
- Kept the existing provider-specific maps/switches rather than doing a representation-only table rewrite. Exact-override enumeration and catalog invariant tests already enforce completeness; a broader refactor is not required for correctness here.

Verification:
- `go test ./...`
- `task lint:new` — 0 issues
- GitHub: license, lint, and tests all passing
- unresolved review threads: 0

malinskibeniamin:
Typed metadata follow-up is complete in 5d7c0d088866.

- ReleaseDate, EndOfLifeDate, and MetadataVerifiedDate are now UTC-midnight time.Time values rather than raw strings.
- Deprecated and Retired were replaced by one zero-safe ModelLifecycle enum: UNKNOWN, ACTIVE, DEPRECATED, or RETIRED.
- EndOfLifeDate remains informational; even a past date never changes Lifecycle without confirmed provider evidence.
- The catalog invariants now require typed dates and an explicit non-unknown lifecycle for every discovered model.
- cloudv2#28132 is updated to consume the enum directly and convert typed dates to google.type.Date without reparsing.

Verification: go test ./...; task license:check; task lint:new with 0 issues. All PR checks pass.

llm/types.go
```diff
@@ -304,6 +307,68 @@ type ModelCapabilities struct {
...
+	Lifecycle ModelLifecycle
+	// ProviderReplacement is an optional migration target named by the
+	// provider. It is not necessarily the latest model and must not be used for
+	// ranking; consumers derive that from UpgradeGroup and ReleaseDate.
+	ProviderReplacement string
+	// OfficialSourceURL identifies the provider evidence for this metadata.
+	OfficialSourceURL string
+	// MetadataVerifiedDate is when a maintainer last checked the metadata
+	// against OfficialSourceURL. It has no lifecycle meaning.
+	MetadataVerifiedDate time.Time
```
weeco, line 350: No one is realistically going to update this all the time. Not sure how this is going to help in the end product consuming the metadata either. Let's remove it.

llm/types.go
```diff
@@ -304,6 +307,68 @@ type ModelCapabilities struct {
...
+	// EndOfLifeDate is the provider-announced retirement schedule, when known.
+	EndOfLifeDate time.Time
+	// Lifecycle records the explicit provider lifecycle statement.
+	Lifecycle ModelLifecycle
+	// ProviderReplacement is an optional migration target named by the
+	// provider. It is not necessarily the latest model and must not be used for
+	// ranking; consumers derive that from UpgradeGroup and ReleaseDate.
+	ProviderReplacement string
+	// OfficialSourceURL identifies the provider evidence for this metadata.
+	OfficialSourceURL string
```
weeco, line 347: Possibly can come from different sources. Why should it be part of the metadata object itself, rather than comments in code where the metadata itself is specified?

llm/types.go
```diff
@@ -304,6 +307,68 @@ type ModelCapabilities struct {
...
+	// ReleaseDate orders versions within a family or upgrade group.
+	ReleaseDate time.Time
+	// EndOfLifeDate is the provider-announced retirement schedule, when known.
+	EndOfLifeDate time.Time
+	// Lifecycle records the explicit provider lifecycle statement.
+	Lifecycle ModelLifecycle
+	// ProviderReplacement is an optional migration target named by the
+	// provider. It is not necessarily the latest model and must not be used for
+	// ranking; consumers derive that from UpgradeGroup and ReleaseDate.
+	ProviderReplacement string
```
weeco, line 345: This is IMO too opinionated to declare. Is gpt frontier the pro models or not for example. Is Opus 5 or Fable 5 better etc.. In my opinion remove it.

llm/types.go
```diff
@@ -304,6 +307,68 @@ type ModelCapabilities struct {
...
+	// UpgradeGroup identifies models whose release dates can be compared to
+	// choose the latest useful version. It is provider-scoped and includes
+	// routing geography where availability differs, such as Amazon Bedrock.
+	UpgradeGroup string
+	// ReleaseDate orders versions within a family or upgrade group.
+	ReleaseDate time.Time
+	// EndOfLifeDate is the provider-announced retirement schedule, when known.
+	EndOfLifeDate time.Time
+	// Lifecycle records the explicit provider lifecycle statement.
+	Lifecycle ModelLifecycle
```
weeco, line 341: What may happen in practice that we set `LifeCycle = Active` and the day it is deprecated or remove the lifecycle still shows Active. So we have double duty here, because the lifecycle is not derived from the date only it seems

llm/types.go
```diff
@@ -304,6 +307,68 @@ type ModelCapabilities struct {
...
+	MetadataVerifiedDate time.Time
+}
+
+// ModelCatalogProvider exposes typed model-catalog facts
+// without changing ModelDiscoveryInfo's public struct layout. Implementations
+// resolve official aliases and dated snapshots to the same metadata as their
+// canonical discovery entry.
+type ModelCatalogProvider interface {
+	ModelCatalog(model string) (ModelCatalogMetadata, bool)
+}
```
weeco, line 359: Hmm if we do this (resolving models from families) it would be good to also return the canoncial model id somewhere so that they know what it was resolved against. Not sure whether this could go to the model catalog metadata or should be a third return param

weeco (changes requested):

PR #178, opened 2026-07-21
anthropic: bill the >200K long-context tier on 1M-window models

birdayz:
## What

Price requests above 200K tokens at Anthropic's long-context rates on the 1M-window Anthropic models. Adds a >200K pricing bracket to Sonnet 5, Opus 4.6/4.7/4.8, and Fable 5 (default and fast-mode cards). Also adds a `model-catalog` skill documenting how pricing is sourced and encoded.

## Why

The pricing engine already models context-size split billing (`TieredInfo`/`Bracket`), and Google (200K) and OpenAI (272K) catalogs use it. Every Anthropic entry was `FlatInfoFromRates`, so a request above 200K on a 1M-window model was billed at the sub-200K flat rate. That undercounts cost on precisely the large requests a 1M window exists to serve.

This started as a task to send the `context-1m-2025-08-07` beta header. That turned out to be a dead end: on the real API the current 1M models (Sonnet 5, Opus 4.6/4.7/4.8, Fable 5) accept >200K requests with no beta header at all — the window is native, the header is a no-op. Sonnet 4.5 is 200K and the header does not lift it. So there is nothing to gate; the only real gap was pricing.

## Implementation details

The surcharge is Anthropic's published structure, uniform across `claude-sonnet-4/4.5/4.6` on the provider page and LiteLLM's `*_above_200k_tokens` fields:

```
input                2x base
output               1.5x base
cache read           2x base
cache write (5m/1h)  2x base
```

Threshold is `MinContextTokens: 200_001`, matching the Google/OpenAI tiers. Each model's default card and any fast-mode override card gets its own bracket (a bracket's `Rates` fully replace the base above the threshold, and an override card carries its own brackets). 200K models are left flat.

LiteLLM's metadata does not yet carry these forward-looking model names, so the brackets are derived from each model's base rates using that uniform multiplier rather than copied from a per-model table. `providers/anthropic/pricing_longcontext_test.go` pins the arithmetic on every card, asserts 200K models grow no bracket, and drives a real `Catalog` end to end to prove a >200K request costs 2x input. If Anthropic ever publishes a model-specific surcharge, that test is where it gets corrected.

The `model-catalog` skill captures the durable process behind this: where to pull rates, the LiteLLM fetch/grep recipe, the microcents convention, flat vs tiered pricing, the long-context surcharge, and how to empirically probe a model's real context window (including the beta-header no-op finding, so nobody repeats the dead end).

## References

- Anthropic pricing: https://docs.anthropic.com/en/docs/about-claude/pricing
- LiteLLM metadata: https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json

### DIFF
--- a/.claude/skills/model-catalog/SKILL.md
+++ b/.claude/skills/model-catalog/SKILL.md
@@ -1,0 +1,132 @@
+---
+name: model-catalog
+description: >-
+  How to source and encode model pricing in this repo's provider catalogs
+  (providers/*/models.go) — where to pull authoritative rates, the microcents
+  convention, flat vs context-tiered pricing (TieredInfo/Bracket), Anthropic's
+  >200K long-context surcharge, cache-rate multipliers, and how to empirically
+  verify a model's real context window. Use when adding or correcting Pricing
+  on a catalog entry, or when a large-context request looks mis-costed. Pairs
+  with the model-maintenance skill and the "Adding New Models" rules in
+  CLAUDE.md.
+---
+
+# Model catalog pricing
+
+Every entry in a provider's `supportedModels` map carries a `Pricing pricing.Info`.
+Getting the numbers right — and the *shape* right (flat vs context-tiered) — is
+the point of this skill. Read the "Adding New Models" section in `CLAUDE.md` for
+the microcents rule; this skill covers where the rates come from and how to encode
+tiers.
+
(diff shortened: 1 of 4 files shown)


alenkacz (approved):
LGTM thanks

PR #179, opened 2026-07-22
fix(agent): deliver truncated responses instead of failing the run

MikolajGagatekRP:
## What

Implements the ai-sdk-go side of RFC [`2607212300-ai-agent-output-truncation`](https://github.com/redpanda-data/cloudv2/pull/28392).

When a managed agent's turn was cut off at the model's `max_tokens` output cap, we **failed the whole run** with *"context length limit exceeded"* — both fatal (wrong) and mislabeled (it was an output-token cut, not a context problem). This PR makes output truncation non-fatal and stops conflating it with genuine context-window overflow, end to end.

## Why

Production incident on a managed agent (claude-sonnet-5): the turn ended `finish_reason: length` at ~103K tokens, far under the window. Three compounding defects, two of which live here:

1. Default output budget hardcoded to `4096` — tiny for an agent, hit routinely.
2. Executor mapped `FinishReasonLength` → `TaskStateFailed` + *"context length limit exceeded"*.
3. Response mapper collapsed `max_tokens` (output truncation) and `model_context_window_exceeded` (input overflow) into one `FinishReasonLength`.

## Changes

- **Split the finish reason end to end.** New `FinishReasonContextOverflow` on both the `llm` and `agent` enums; the provider stop reason for input overflow maps to it, while output truncation stays `FinishReasonLength`. Threaded `llm` → `agent` → executor.
- **Executor: truncation is `completed`, not `failed`** (`adapter/a2a/executor.go`). Output truncation emits `TaskStateCompleted` with a `truncated: true` marker on the **message** metadata (where the UI reads it) plus a truthful notice; the partial content is delivered. Genuine context overflow stays terminal with an accurate, actionable message. Continuation isn't offered for overflow because the input is already at the window ceiling — a new turn would immediately re-overflow.
- **Output budget: reframed 16K fallback + per-request override.** `defaultMaxTokens` (16K) is now documented as a last-resort fallback to satisfy Anthropic's required field, **not** a recommended budget — the budget *policy* belongs to the harness. Added `anthropic.RequestOptions{MaxTokens}`, read from `llm.Request.Options`, so a harness can set `max_tokens` per turn (clamped to `MaxOutputTokens`) without rebuilding the model. Precedence: per-request → `WithMaxTokens` → fallback.
- **Adapters keep the wire value stable** (`plugins/otel/transform.go`, `adapter/vercelaisdk/uimessagestream/handler.go`): explicit `FinishReasonContextOverflow` → `"length"` arms so the split doesn't silently reclassify overflow.

## Provider coverage (the finish-reason split)

The split matters only where overflow arrives as a **stop reason** rather than a 400 error:

| Provider | Output truncation | Context overflow | Status |
|---|---|---|---|
| Anthropic | `max_tokens`→Length | `model_context_window_exceeded`→ContextOverflow | ✅ fixed |
| Bedrock | `MaxTokens`→Length | `ModelContextWindowExceeded`→ContextOverflow | ✅ fixed |
| Google / OpenAI / openaicompat | truncation→Length | 400 `context_length_exceeded` → executor error path | ✅ already truthful, no change needed |

## Tests (TDD — written first, watched fail)

- `providers/anthropic` + `providers/bedrock`: `max_tokens` vs `model_context_window_exceeded` map to distinct reasons.
- `providers/anthropic`: per-request override wins / clamps to `MaxOutputTokens` / falls back; default budget > 4096.
- `providers/anthropic`: **e2e integration test** against the real Anthropic API — a 16-token per-request cap truncates a long turn to exactly 16 output tokens (overriding a per-model 4096), a generous cap completes normally.
- `agent/llmagent`: context overflow surfaces as `agent.FinishReasonContextOverflow`.
- `adapter/a2a`: output truncation → `Completed` + `truncated` marker + partial delivered; overflow → `Failed` with a truthful, non-misleading message.
- `adapter/vercelaisdk`: overflow maps to `"length"`.

`task lint:new` → 0 issues; `task license:check` clean; full `go test -short ./...` green; CI green (incl. the e2e integration test).

## Differences from the RFC

This PR is the ai-sdk-go slice only, and deviates from the RFC in a few places:

**Beyond the RFC:**
- The RFC named only the Anthropic response mapper for the split; this PR also splits **Bedrock** (it has the same stop reason and the same conflation).
- Added explicit **otel + vercelaisdk adapter** handling — the RFC didn't mention these, but the split would otherwise silently change their wire value (overflow → `"other"`/`"error"`).
- Added a **per-request `RequestOptions.MaxTokens` override**. The RFC only anticipated per-model `With*` options plus a future dynamic cap; per-maintainer review the budget policy belongs to the harness, so this is the mechanism that lets the harness own it.
- Chose to place the raised default in **ai-sdk-go** (the RFC left it "ai-sdk-go or ai-agent").

**Deferred to follow-ups / other repos (as the RFC scopes them):**
- **Dynamic per-request cap** (`min(MaxOutputTokens, window − est_input − margin)`) — RFC's preferred end state; belongs in apps/ai-agent, not the SDK.
- **apps/ai-agent**: persist the partial turn + marker, per-turn finish/stop-reason + token logging, catalog-driven generation config.
- **apps/adp-ui**: read the `truncated` marker, truncation notice + **Continue** action, error-card branching.

**Partial vs the RFC:**
- **Metadata placement**: the RFC flagged that the executor writes `finish_reason`/`usage` on the status-*event* metadata while adp-ui reads off the status-*message* metadata. This PR puts the new `truncated` signal on the **message** metadata (the part adp-ui keys on); `usage` is also already delivered on each message-event's metadata. The legacy `finish_reason`/`usage` on the terminal *event* metadata is left as-is.
- **State 2 (pre-generation `input + max_tokens > window` 400)**: surfaces through the executor's generic error path with the provider's raw error string, not a custom truthful message. The truthful-message treatment is implemented for the `model_context_window_exceeded` *stop-reason* path (`FinishReasonContextOverflow`).

### DIFF
--- a/llm/types.go
+++ b/llm/types.go
@@ -274,8 +274,19 @@ const (
 	// FinishReasonStop indicates the model completed naturally.
 	FinishReasonStop FinishReason = "stop"
 
-	// FinishReasonLength indicates the response was truncated due to length limits.
+	// FinishReasonLength indicates the response was truncated at the output-token
+	// cap. This is a non-fatal, per-turn truncation: the model produced a partial
+	// response and stopped. Distinct from FinishReasonContextOverflow.
 	FinishReasonLength FinishReason = "length"
+
+	// FinishReasonContextOverflow indicates generation stopped because the
+	// conversation exceeds the model's context window (the provider's
+	// context-window stop reason, e.g. Anthropic's model_context_window_exceeded),
+	// as opposed to FinishReasonLength which is a per-turn output-token cut.
+	// Terminal: the conversation must be shortened or restarted, not continued.
+	// Any content the model produced before the limit is still delivered on the
+	// response — this reason marks the cause, it does not imply empty content.
+	FinishReasonContextOverflow FinishReason = "context_overflow"
 
 	// FinishReasonToolCalls indicates the model wants to execute tools.
 	FinishReasonToolCalls FinishReason = "tool_calls"
--- a/providers/anthropic/provider.go
+++ b/providers/anthropic/provider.go
@@ -167,6 +167,21 @@ func WithTimeout(timeout time.Duration) ProviderOption {
 	}
 }
 
+// defaultMaxTokens is a last-resort fallback, NOT a recommended budget.
+//
+// Anthropic requires max_tokens on every request, so the SDK must send some
+// value when neither WithMaxTokens nor a per-request override (RequestOptions)
+// is set. The budget *policy* belongs to the caller, not the SDK — so this value
+// only needs to be a safe default that works out of the box, not the "right"
+// number for any given workload.
+//
+// It is deliberately bounded rather than the model's output max: max_tokens is a
+// reservation against the context window (input + max_tokens must fit), so
+// defaulting to the model max would 400 long conversations. 16K is generous but
+// bounded: enough for long answers, small enough to stay clear of context-window
+// rejections on typical conversations.
+const defaultMaxTokens = 16384
+
 // NewModel creates a new Anthropic model instance with the specified configuration.
 func (p *Provider) NewModel(modelName string, opts ...Option) (llm.Model, error) {
 	family := resolveModelFamily(modelName)
@@ -179,7 +194,7 @@ func (p *Provider) NewModel(modelName string, opts ...Option) (llm.Model, error)
 	cfg := &Config{
 		ModelName:        modelName,
 		Constraints:      modelDef.Constraints,
-		MaxTokens:        4096, // Default required by Anthropic API
+		MaxTokens:        defaultMaxTokens, // Required by Anthropic API; see const.
 		EnableCaching:    p.EnableCaching,
 		AdaptiveThinking: modelDef.AdaptiveThinking,
 		setOptions:       make(map[string]bool),
--- a/adapter/a2a/executor.go
+++ b/adapter/a2a/executor.go
@@ -247,9 +247,30 @@ func (e *Executor) processEvents(
 					Text: "Agent stopped: maximum iterations reached",
 				})
 			case agent.FinishReasonLength:
+				// Output truncation is non-fatal: the model produced a partial
+				// response and stopped at the output-token cap. Complete the task,
+				// deliver what we have (already streamed to history), and mark the
+				// turn truncated so the surface can offer a Continue action instead
+				// of a destructive failure card.
+				taskState = a2a.TaskStateCompleted
+				statusMsg = a2a.NewMessage(a2a.MessageRoleAgent, a2a.TextPart{
+					Text: "Response was truncated — the maximum output token limit was reached. Continue to get the rest.",
+				})
+				// The `truncated` marker is the load-bearing contract: it lives on
+				// the message metadata (where consumers read usage), NOT the event
+				// metadata. This notice is a UI affordance, not model output — a
+				// consumer that reconstructs conversation history from agent status
+				// messages MUST skip messages carrying `truncated: true`, otherwise a
+				// "Continue" would resume from a history where the assistant appears
+				// to have said "Continue to get the rest." The partial answer was
+				// already delivered as a separate Working-status MessageEvent.
+				statusMsg.Metadata = map[string]any{"truncated": true}
+			case agent.FinishReasonContextOverflow:
+				// Genuinely too long: the input exceeded the model's context window,
+				// so nothing could be generated. Terminal, but say so truthfully.
 				taskState = a2a.TaskStateFailed
 				statusMsg = a2a.NewMessage(a2a.MessageRoleAgent, a2a.TextPart{
-					Text: "Agent stopped: context length limit exceeded",
+					Text: "Agent stopped: the conversation exceeds the model's context window. Start a new conversation or shorten the input.",
 				})
 			case agent.FinishReasonError:
 				taskState = a2a.TaskStateFailed
--- a/adapter/a2a/truncation_test.go
+++ b/adapter/a2a/truncation_test.go
@@ -1,0 +1,198 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package a2a
+
+import (
+	"context"
+	"log/slog"
+	"strings"
+	"testing"
+
+	"github.com/a2aproject/a2a-go/a2a"
+	"github.com/a2aproject/a2a-go/a2asrv"
+	"github.com/a2aproject/a2a-go/a2asrv/eventqueue"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/redpanda-data/ai-sdk-go/agent/llmagent"
+	"github.com/redpanda-data/ai-sdk-go/llm"
+	"github.com/redpanda-data/ai-sdk-go/llm/fakellm"
+	"github.com/redpanda-data/ai-sdk-go/runner"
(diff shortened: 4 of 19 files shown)


MikolajGagatekRP:
@claude review

MikolajGagatekRP:
Thanks for the review. Addressed in `8bb3fb1`:

- **#.1 (other `FinishReason` consumers)** — Fixed. Added explicit `case llm.FinishReasonContextOverflow` to both `adapter/vercelaisdk/uimessagestream/handler.go` and `plugins/otel/transform.go`, mapping to `"length"` (the token-limit bucket, which is also what overflow wired to before the split — so no silent behavior change). Extended the Vercel mapping test.
- **#.2 (notice polluting continuation history)** — Documented the contract explicitly in `executor.go`: the `truncated: true` marker is load-bearing, and consumers reconstructing history from agent status messages must skip messages carrying it. Kept the notice text in content (the exact message shape is a contract being settled with the adp-ui side of the RFC; didn't want to guess it here).
- **#.3 (agent-as-tool truncation)** — Agreed, out of scope. Will file a follow-up ticket if managed agents start composing sub-agents.

Nits confirmed, no action. Full `go test -short ./...` green, `task lint:new` 0 issues.

MikolajGagatekRP:
@claude review

llm/types.go
```diff
@@ -274,9 +274,17 @@ const (
...
-	// FinishReasonLength indicates the response was truncated due to length limits.
+	// FinishReasonLength indicates the response was truncated at the output-token
+	// cap. This is a non-fatal, per-turn truncation: the model produced a partial
+	// response and stopped. Distinct from FinishReasonContextOverflow.
 	FinishReasonLength FinishReason = "length"
 
+	// FinishReasonContextOverflow indicates the request could not be generated
+	// because the input (optionally plus the reserved output budget) exceeds the
+	// model's context window. Unlike FinishReasonLength, no useful partial
+	// response was produced; the conversation is genuinely too long.
```
weeco: @claude Is it true that no partial response will be returned in that case for the given providers?

providers/anthropic/provider.go
```diff
@@ -167,6 +167,17 @@ func WithTimeout(timeout time.Duration) ProviderOption {
...
+// defaultMaxTokens is the default output-token budget when the caller sets none.
+//
+// Anthropic requires max_tokens on every request. It is a reservation against
+// the context window (input + max_tokens must fit), not just a spend ceiling, so
+// the model max is the wrong default: it 400s long conversations. The old 4096
+// was the opposite footgun — it truncated ordinary agent turns (AI-1674). 16K is
+// generous but bounded: enough for long answers, small enough to stay clear of
+// context-window rejections on typical conversations, matching the bound serious
+// agents such as charmbracelet/crush pick for large Sonnet models.
+const defaultMaxTokens = 16384
```
weeco: Instead of hardcoding, we should probably:

1. Default to the max output token of the respective model
2. Make it possible to override the max tokens for users on a per-request basis.

Raising it from the hardcoded 4k is an improvement already, but we shouldnt hardcode it

providers/anthropic/provider.go
```diff
@@ -167,6 +167,17 @@ func WithTimeout(timeout time.Duration) ProviderOption {
...
 
+// defaultMaxTokens is the default output-token budget when the caller sets none.
+//
+// Anthropic requires max_tokens on every request. It is a reservation against
+// the context window (input + max_tokens must fit), not just a spend ceiling, so
+// the model max is the wrong default: it 400s long conversations. The old 4096
+// was the opposite footgun — it truncated ordinary agent turns (AI-1674). 16K is
+// generous but bounded: enough for long answers, small enough to stay clear of
+// context-window rejections on typical conversations, matching the bound serious
+// agents such as charmbracelet/crush pick for large Sonnet models.
```
weeco: Update comment to not include internal jira references or other repositories from research. Keep it open source friendly, keeping comment concise / limited to the relevant parts

providers/anthropic/provider.go
birdayz: agree. if at all, this choice is up to the harness, e.g. the ai-agent. would not do it here.

providers/anthropic/provider.go
MikolajGagatekRP: sure didnt notice that one

providers/anthropic/provider.go
MikolajGagatekRP: Agreed on both counts, and @birdayz's point that the budget policy belongs in the harness is the right framing. Addressed in `2da1bff`:

- **Per-request override (weeco #.2):** added `anthropic.RequestOptions{MaxTokens *int}`, read from `llm.Request.Options` in the request mapper. A harness can now set `max_tokens` per turn (clamped to the model's `MaxOutputTokens`) without rebuilding the model. Precedence: per-request → `WithMaxTokens` → fallback.
- **Not hardcoding policy:** reframed `defaultMaxTokens` as an explicit *last-resort fallback* to satisfy Anthropic's required field, not a recommended budget. Doc comment now says the budget choice is the caller/harness's job.

On **#.1 (default to the model max)**: keeping it bounded rather than the model max on purpose — `max_tokens` reserves context-window space, so defaulting Sonnet-5 to its 128K output max 400s any conversation with >72K input (`input + max_tokens > window`), which is the exact long-conversation failure this PR fixes (RFC lists it under rejected alternatives). With the per-request override in place, the harness/ai-agent can compute the model-aware budget (dynamic cap or catalog-driven), which is where that logic belongs per @birdayz. Left the dynamic cap out of the SDK accordingly.

llm/types.go
MikolajGagatekRP: Good catch — the comment overstated it, fixed in `ddaadbc`.

No, it's not guaranteed that there's no partial content. `model_context_window_exceeded` is a **stop reason on a 200 response** (like `max_tokens`), returned when the window is hit *during* generation, so the model may have produced some output before stopping. My comment claiming "no useful partial response was produced" was wrong; I've reworded it to say the reason marks the *cause* (conversation exceeds the window → terminal, must be shortened not continued), not that the turn is empty.

Important: this is a **doc fix only, no behavior change**. The agent loop emits the `MessageEvent` with the full `resp.Message` *before* the finish-reason check (`llmagent.go:298` vs `:307`), so whatever partial content exists is delivered to the executor as a working-status message regardless of whether the reason is `Length` or `ContextOverflow`. The split only changes the *terminal* A2A status (truncation → `Completed` + `truncated` marker; overflow → `Failed` + truthful message) — it never gates content delivery, so nothing is dropped.

For Bedrock the same holds (`ModelContextWindowExceeded` is likewise a stop reason). Google/OpenAI surface overflow as a 400 instead, which flows through the executor's error path.

providers/anthropic/provider.go
```diff
@@ -167,6 +167,22 @@ func WithTimeout(timeout time.Duration) ProviderOption {
 	}
 }
 
+// defaultMaxTokens is a last-resort fallback, NOT a recommended budget.
+//
+// Anthropic requires max_tokens on every request, so the SDK must send some
+// value when neither WithMaxTokens nor a per-request override (RequestOptions)
+// is set. The budget *policy* belongs to the caller/harness (e.g. apps/ai-agent),
+// not the SDK — so this value only needs to be a safe default that works out of
+// the box, not the "right" number for any given workload.
```
weeco: References outside this repository shouldnt live here.

MikolajGagatekRP: fixed and also verified if not any more of such in there

providers/anthropic/provider.go
```diff
@@ -167,6 +167,22 @@ func WithTimeout(timeout time.Duration) ProviderOption {
...
+// Anthropic requires max_tokens on every request, so the SDK must send some
+// value when neither WithMaxTokens nor a per-request override (RequestOptions)
+// is set. The budget *policy* belongs to the caller, not the SDK — so this value
+// only needs to be a safe default that works out of the box, not the "right"
+// number for any given workload.
+//
+// It is deliberately bounded rather than the model's output max: max_tokens is a
+// reservation against the context window (input + max_tokens must fit), so
+// defaulting to the model max would 400 long conversations. The old 4096 was the
+// opposite footgun — it truncated ordinary agent turns. 16K is generous but
```
weeco: we dont need to comment about previous values

weeco (approved):

providers/anthropic/provider.go
MikolajGagatekRP: right aldo addressed that

PR #180, opened 2026-07-23
feat(providers): support provider-native thinking controls

malinskibeniamin:
## What changed

- add Gemini 3 thinking levels with exact model-level validation
- preserve Gemini 2.5 token budgets, validate their model-specific ranges, and make the budget option enable thinking
- add Bedrock adaptive thinking and effort controls with exact Claude model/mode validation
- preserve manual Bedrock thinking budgets on supported Claude models
- preserve signature-only adaptive reasoning blocks across streaming and multi-turn round trips
- cover serialized provider payloads and exercise Bedrock adaptive thinking against the live provider

## Why

Gemini 3 uses `thinkingLevel`, not `thinkingBudget`. Adaptive-only Claude models on Bedrock require `thinking: {type: "adaptive"}` plus `output_config.effort` and reject the legacy manual-budget shape. Newer models can omit readable reasoning while still returning a signature; retaining that signature is required for multi-turn continuity.

This is the SDK-first change needed before wiring the controls into [cloudv2#28364](https://github.com/redpanda-data/cloudv2/pull/28364).

## Verification

- `task license:check`
- `task lint:new`
- `task test`
- `go build ./...`
- `go vet ./...`
- GitHub: license, lint, and full provider tests passed

## Provider references

- [Gemini thinking](https://ai.google.dev/gemini-api/docs/generate-content/thinking)
- [Bedrock adaptive thinking](https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html)
- [Claude adaptive thinking](https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking)
- [Claude effort](https://platform.claude.com/docs/en/build-with-claude/effort)


simon0191 (approved):

malinskibeniamin:
@claude review

MikolajGagatekRP (approved):
lgtm

weeco:
## Proposal: one option name for reasoning effort across all providers

We reworked the thinking controls on top of this branch so that every provider exposes the same knob with the same name:

```go
openai.WithReasoningEffort(openai.ReasoningEffortHigh)       // → reasoning.effort
anthropic.WithReasoningEffort(anthropic.ReasoningEffortHigh) // → output_config.effort
bedrock.WithReasoningEffort(bedrock.ReasoningEffortHigh)     // → thinking: adaptive + output_config.effort
google.WithReasoningEffort(google.ReasoningEffortHigh)       // → thinkingLevel: "HIGH"
```

### What changed vs this PR, and why

**`bedrock.WithAdaptiveThinking(EffortLow)` → `bedrock.WithReasoningEffort(ReasoningEffortLow)`.**
"Adaptive thinking" is an Anthropic wire mechanism, not a user intent — and Bedrock also hosts non-Anthropic models (gpt-5.x via mantle) where the very same knob must map to OpenAI's `reasoning.effort` instead. The option names what the user wants; the request mapper picks the wire format per model. It also matches the native `anthropic` provider, so switching a Claude model between the two providers no longer means learning a different API.

**`google.WithThinkingLevel(ThinkingLevelLow)` → `google.WithReasoningEffort(ReasoningEffortLow)`.**
Same reasoning, plus this removes `type ThinkingLevel = genai.ThinkingLevel` — we don't want a third-party type in our public API surface. We map to Gemini's uppercase enum internally.

**Bedrock capabilities moved from a switch into the catalog.**
`modelThinkingCapabilities()` is replaced by a `Thinking ThinkingSupport` field on each `ModelDefinition` entry (shared per-generation vars, same pattern as `claudeStandardCaps`). A new model's thinking support now lives next to all its other metadata, like the openai/anthropic catalogs already do.

**`anthropic.WithEffort` → `anthropic.WithReasoningEffort`** so the name is identical everywhere (openai already used it).

### The type is shared, the values are not

The `llm` package defines only plumbing — deliberately **no constants**:

```go
// llm
type ReasoningEffort string

type ReasoningEffortLister interface {
    SupportedReasoningEfforts() []ReasoningEffort
}
```

Effort vocabulary is provider-owned. Each provider declares typed constants for exactly the values its models accept, so autocomplete on `google.` only offers `Minimal/Low/Medium/High`, and a future provider with its own vocabulary ("intense", numeric scales, …) just declares it locally — `llm` never changes.

Which subset a *specific model* supports stays catalog metadata, enforced at construction with a self-explaining error:

```
model claude-sonnet-4-6 does not support reasoning effort "max" (supported: [low medium high])
```

Multi-provider consumers (e.g. cloudv2) can discover valid values without type-switching on concrete providers:

```go
if lister, ok := model.(llm.ReasoningEffortLister); ok {
    efforts := lister.SupportedReasoningEfforts() // e.g. populate a settings dropdown
}
```

### Unchanged from this PR

Manual token budgets stay provider-specific (`google.WithThinkingBudget`, `bedrock.WithThinking(budget)`) with this PR's range and model validation; the signature-only reasoning round-trip is untouched. Combining an effort with a manual budget is rejected.

The rework is implemented and green (`task lint:new`, license check, full unit suite) on a local branch stacked on this one — happy to push it here or open it as a follow-up PR, whichever you prefer.

weeco:
Pushed the rework described above to this branch as c929bdf — lint:new, license check, and the full unit suite are green. The Bedrock/Google live integration tests still deserve one pass with real credentials since the adaptive wire path changed.

ZK-RPD:
Picking this up under [AI-1539](https://redpandadata.atlassian.net/browse/AI-1539) following Ben's handover. Merged `main` in and taken this out of draft — it was 16 commits behind, and Claude Opus 5 support plus the GPT-5.6 model family landed in the meantime.

### Conflict resolution

**`providers/bedrock/models.go`** — one file, four regions:

- The first was purely additive: this branch adds the `ReasoningEffort` alias and `ThinkingSupport` catalog metadata, `main` adds the GPT-5.6 model ID constants. Kept both.
- The three Fable 5 entries (global/US/EU) now use `main`'s `claudeNoSampling1MConstraints` — the rename of `claudeFable5Constraints`, which also drops `TemperatureRange` since those models reject sampling params — while retaining this branch's `Thinking: frontierClaudeThinking`.

**`providers/anthropic`** — two semantic conflicts that merged cleanly but broke the build, since `main`'s new Opus 5 entry was written against the pre-rename option surface:

- `SupportedEfforts: []Effort{...}` → `SupportedReasoningEfforts: []ReasoningEffort{...}`
- `alias_test.go`: `WithEffort` → `WithReasoningEffort`, `config.Effort` → `config.ReasoningEffort`, `EffortLow..EffortMax` → `ReasoningEffort*`

### One real defect fixed

`main`'s Opus 5 entry declared:

```go
SupportedParams: []string{"max_tokens", "effort", "speed"}
```

but `ValidateParameterSupport` and every other model in the catalog use `"reasoning_effort"`. So `WithReasoningEffort` on Opus 5 failed with `parameter reasoning_effort is not supported`. Corrected to `"reasoning_effort"`.

Worth flagging on its own: this would have shipped with effort controls silently broken on Opus 5.

### Verification

`go build ./...` and `go vet ./...` clean. `llm`, `llm/fakellm` and all five provider packages pass. The only repo-wide failures are the two `kvstore` packages, which need Docker (testcontainers) and are unrelated to this change — CI will cover those.

### Notes for review

- The approval predates the merge, so a re-look at the resolution would be welcome — particularly the Fable 5 constraints change, since dropping `TemperatureRange` is a behaviour change inherited from `main` rather than something this branch chose. Related: [AI-1693](https://redpandadata.atlassian.net/browse/AI-1693) (Gemini sampling deprecation) covers the same theme.
- Not merging yet — running a review pass first.
- cloudv2#28364 is the consumer half and still needs review; it'll want a rebase once this lands.


[AI-1539]: https://redpandadata.atlassian.net/browse/AI-1539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AI-1693]: https://redpandadata.atlassian.net/browse/AI-1693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ZK-RPD:
CI status: **License Headers pass, Lint pass, Test fails — but the failure is pre-existing and unrelated to this branch.**

The only failing test is the live-API conformance test:

```
TestAnthropicConformance_Integration/TestAllSupportedModels/
  basic_generation_works_for_all_supported_models/model_claude-opus-4-1

POST https://api.anthropic.com/v1/messages: 404 Not Found
{"type":"not_found_error","message":"model: claude-opus-4-1"}
```

`claude-opus-4-1` has been retired upstream. Evidence it isn't this branch:

- `main`'s last Test run passed on 2026-08-04; the model 404s as of today. `main` would fail this identically if it ran now — it just hasn't since the 4th.
- This merge doesn't touch the `ModelClaudeOpus41` entry (context-only in the diff).
- Every non-integration anthropic test passes, including all the effort/thinking validation.

Two other integration tests are also environment-gated rather than broken: Bedrock Fable 5 skips on an AWS Marketplace subscription (`AccessDeniedException`), and the Google suite is disabled by default behind `RUN_GOOGLE_INTEGRATION=1`.

So the catalog needs `claude-opus-4-1` retired (or the conformance list narrowed) — worth its own ticket rather than widening this PR's scope. Happy to file it. Flagging so the red X isn't read as a regression here.

ZK-RPD:
Refreshed against `main` now that #192 has landed. The `claude-opus-4-1` failure is gone, and Lint and License Headers pass.

**Test is still red, but on a different and again unrelated cause: two OpenAI models have gone missing upstream today.**

```
TestOpenAIConformance_Integration/TestAllSupportedModels/
  basic_generation_works_for_all_supported_models/model_gpt-5.2-chat-latest
TestAllModelsReasoningEffortsIntegration/gpt-5.2-chat-latest/medium
  (and the gpt-5.3-chat-latest equivalents)

API call failed: Model not found gpt-5.2-chat-latest
API call failed: Model not found gpt-5.3-chat-latest
```

### This is not the branch

- `main`'s Test run for the #192 merge ([31584599870](https://github.com/redpanda-data/ai-sdk-go/actions/runs/31584599870), 09:48 UTC today) **passed these exact subtests** — `model_gpt-5.2-chat-latest`, `model_gpt-5.3-chat-latest`, and both `/medium` effort cases. This branch failed them at 15:31 UTC, so they became unavailable in between.
- `git diff main..HEAD -- providers/openai/` is limited to making `ReasoningEffort` an alias of `llm.ReasoningEffort`, returning a cloned slice from `SupportedReasoningEfforts()`, and adding the supported list to one error string. Nothing in it changes the model name sent to the API, so it cannot produce a `Model not found` from OpenAI.
- The model catalogs are byte-identical between `main` and this branch.

`main` would fail the same way if it ran now.

### Also in this push

One genuine fix of mine: the `SupportedEfforts` → `SupportedReasoningEfforts` rename in the earlier merge widened the struct's alignment column, and I left `SupportedSpeeds` / `AdaptiveThinking` on their old padding in the Opus 5 entry. `gofmt` flagged it; realigned in `d39088d`. (`providers/openaicompat/openaicompat_test.go` is also unformatted, but it is that way on clean `main` too — not touching it here.)

### Where that leaves this PR

Same shape as AI-1779, one provider over. Options:

1. Retire `gpt-5.2-chat-latest` and `gpt-5.3-chat-latest` from the OpenAI catalog in a separate PR, exactly as #192 did, then this goes green.
2. Merge this on the strength of the evidence above, and retire them separately.

I'd rather not widen this PR to include an unrelated catalog change. Flagging for a call — and it is worth noting this is the second upstream retirement in three days, so a periodic catalog audit (or a conformance path that reports retirement without blocking) may be worth its own ticket.

ZK-RPD:
## Merging with a red Test check — evidence that this branch is not the cause

Lint and License Headers pass. Test fails on **two OpenAI models only**, and the failure is provider-side. Setting out the evidence in full so this is unambiguous for anyone who finds a red `main` later.

### The failure

```
API call failed: Model not found gpt-5.2-chat-latest
API call failed: Model not found gpt-5.3-chat-latest
```

`Model not found` is OpenAI's API rejecting the model identifier. Reproduced on re-run, so it is not transient.

### Evidence, five independent lines

**1. `main` passed these exact subtests today, hours earlier.**
Run [31584599870](https://github.com/redpanda-data/ai-sdk-go/actions/runs/31584599870) (the #192 merge commit, 09:48 UTC 2026-08-12) passed `model_gpt-5.2-chat-latest`, `model_gpt-5.3-chat-latest`, `TestAllModelsReasoningEffortsIntegration/gpt-5.2-chat-latest/medium` and the 5.3 equivalent. This branch failed the same four at 15:31 UTC. They became unavailable in between.

**2. Only the `-chat-latest` aliases fail. The generation works.**
From the same failing run:

| Model | Result |
|---|---|
| `gpt-5.2` | PASS |
| `gpt-5.2-pro` | PASS |
| `gpt-5.2-chat-latest` | **FAIL** |
| `gpt-5.3-chat-latest` | **FAIL** |

21 other OpenAI models pass in that run. No change in this branch is alias-selective — nothing here could break `gpt-5.2-chat-latest` while leaving `gpt-5.2` and `gpt-5.2-pro` working.

**3. The OpenAI model catalogs are byte-identical between `main` and this branch.**
`diff` of the model-ID constants in `providers/openai/models.go` is empty. This branch does not add, rename or repoint a single OpenAI model.

**4. The entire `providers/openai` diff cannot produce a `Model not found`.**
It is three things: `ReasoningEffort` becoming a type alias of `llm.ReasoningEffort`, `SupportedReasoningEfforts()` returning `slices.Clone(...)`, and one error string gaining the supported-values list. None of them touches the model name sent on the wire.

**5. The failing tests already existed on `main`.**
`TestAllModelsReasoningEffortsIntegration` and the OpenAI conformance cases are not introduced here — they are present on `main` and passed there this morning (see 1).

### Consistent with documented repo behaviour

`CLAUDE.md` states: *"Integration tests against LLM providers can be flaky due to provider behavior."* These are live-provider integration tests. The suite already has `isModelAccessGate`, which exists precisely so provider-side availability skips rather than fails conformance — the two shapes it covers are Anthropic gated models and Bedrock models without a Marketplace subscription. A bare OpenAI `Model not found` is the same class of condition and is not yet covered by that matcher.

### Deliberately not fixing it here

Removing or gating these two aliases is an unrelated OpenAI catalog change, and widening an approved PR to carry it would be the wrong call. It is also not yet established whether OpenAI has withdrawn the aliases or the CI account's model access changed — those need different fixes, so I am recording the state as *unavailable to CI as of 2026-08-12* rather than asserting retirement. Tracked separately; ticket reference to follow in this thread.

### Also in this push

One genuine fix of my own: the `SupportedEfforts` → `SupportedReasoningEfforts` rename in the earlier merge widened the struct's alignment column and left `SupportedSpeeds` / `AdaptiveThinking` on their old padding in the Opus 5 entry. `gofmt` flagged it; realigned in `d39088d`. (`providers/openaicompat/openaicompat_test.go` is also unformatted, but is so on clean `main` — untouched here.)

### For the record

Also worth noting this branch is in better shape than at handover: the five Bedrock reasoning failures Blacksmith reported on 2026-07-23 all pass now, and every one of the 10 failures in the current Blacksmith report is one of the two `-chat-latest` aliases. Nothing else in the suite is red.

PR #185, opened 2026-07-28
feat(llm,anthropic,a2a): surface context-window overflow as a typed terminal error

MikolajGagatekRP:
## What

Follow-up to #179. That change split output truncation (non-fatal) from context-window overflow when the model **stops on a 200 response** (`FinishReasonContextOverflow`). This closes the other half: when the provider **rejects the request before generating** because the conversation no longer fits the context window.

- New sentinel `llm.ErrContextWindowExceeded`.
- The Anthropic error classifier detects the pre-generation 400 — `"input length and max_tokens exceed context limit: N + M > W"` (input + max_tokens) and `"prompt is too long: …"` (input alone) — on both the HTTP and streaming error paths, and wraps it in the new sentinel. Other bad requests stay `ErrInvalidInput`.
- The A2A executor maps the typed error to a **terminal failure** carrying the same truthful, actionable message it already uses for `FinishReasonContextOverflow`, instead of leaking the raw provider error string.

## Why

The two conditions are the same thing from the user's point of view — "the conversation is too long" — but they arrive on different code paths (a 200 finish reason vs. a pre-generation 400). Before this change only the 200 path was told apart; the 400 fell into generic invalid-request handling and surfaced an opaque error. Now the outcome is consistent whichever path fires.

## Testing

- `providers/anthropic`: classifier unit tests for both overflow messages on the HTTP and SSE paths, plus a guard that an ordinary `invalid_request_error` stays `ErrInvalidInput`.
- `adapter/a2a`: executor test proving the typed error becomes a terminal failure with the actionable message.
- Built with TDD (red → green for each).

## Scope / follow-ups

- **Anthropic only.** Detection uses Anthropic's confirmed 400 strings. OpenAI/Google/Bedrock can adopt the same sentinel once their exact rejection messages are confirmed — deferred rather than guessed.
- **Budget sizing stays with the caller.** A dynamic `max_tokens` clamp (size the request so it never triggers this 400) is budget policy and belongs in the calling application, which already has the per-request override hook from #179. This PR only makes the SDK *classify the rejection truthfully*; it does not size requests.

### DIFF
--- a/adapter/a2a/context_overflow_error_test.go
+++ b/adapter/a2a/context_overflow_error_test.go
@@ -26,10 +26,11 @@ import (
 )
 
 // TestExecutor_ContextWindowExceededIsTerminalAndTruthful verifies that when the
-// provider rejects the request before generation because the conversation does
-// not fit the context window (surfaced as llm.ErrContextWindowExceeded), the
-// executor fails the task with the same truthful, actionable message it uses for
-// the 200-response FinishReasonContextOverflow — not the raw provider error.
+// provider rejects the request before generation because it does not fit the
+// context window (surfaced as llm.ErrContextWindowExceeded), the executor fails
+// the task with a truthful, actionable message — not the raw provider error. The
+// rejection also covers input + max_tokens overflowing while the input alone
+// fits, so the message must name lowering the response limit too.
 func TestExecutor_ContextWindowExceededIsTerminalAndTruthful(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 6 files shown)


MikolajGagatekRP:
@claude review

malinskibeniamin (commented):
Automated `/review`: No P0/P1 findings.

malinskibeniamin (commented):
Automated `/review`: No P0/P1 findings.

malinskibeniamin (approved):

PR #186, opened 2026-07-28
feat(agenttool): map sub-agent terminal states (truncation, overflow, errors)

MikolajGagatekRP:
## What

When an agent is used as a tool (`tool/agenttool`), `Execute` captured the sub-agent's last message text but ignored its finish reason — so **every** terminal state (truncation, context overflow, content-filter/unknown errors, max-turns, input-required) came back as an indistinguishable success with whatever partial text existed.

This branches on the sub-agent's terminal finish reason (and surfaces a fatal `ErrorEvent` cause), mirroring the A2A executor so both composition surfaces agree on success vs failure:

| Sub-agent finish reason | agent-as-tool outcome |
|---|---|
| Stop / Transfer | success (`{"result": "..."}`) |
| **Length** (output truncation) | success + `{"metadata": {"truncated": true}}`, partial content delivered |
| ContextOverflow | tool error ("conversation exceeds the model's context window") |
| Error (content filter / unknown) | tool error, surfacing the underlying `ErrorEvent.Err` |
| MaxTurns | tool error ("maximum iterations reached") |
| InputRequired | tool error (a tool can't pause for external input) |

`metadata` is keyed the same way as the A2A executor's status-message metadata, so both surfaces share one vocabulary; it's the extension point for further markers (e.g. `usage`).

Genuine generation/transport errors (incl. the pre-generation context-window 400) already arrive via the iterator error channel and are propagated as a tool error — unchanged.

## Why

`llmagent` signals most fatal conditions through the **finish reason** / a non-terminal `ErrorEvent` (both yielded with a nil iterator error), not through `yield(nil, err)`. `agenttool` previously only checked the iterator error + last message, so those slipped through as successes. The A2A executor already detects fatality from the finish reason; this brings agent-as-tool to parity.

## Testing

- Unit (`TestExecute`): truncation → `metadata.truncated` + partial; natural stop → no metadata; context-overflow / max-turns / input-required → tool error; terminal error surfaces the underlying cause (`ErrContentPolicyViolation`).
- E2E integration: real `llmagent` on the live Anthropic model with a 16-token cap genuinely truncates → `metadata.truncated=true`. Key-gated.
- Built TDD (red → green).

## Notes

Absorbs the follow-up that was filed as AI-1711 (agent-as-tool swallowing overflow/error terminal states) — handled here rather than deferred.

### DIFF
--- a/tool/agenttool/agenttool.go
+++ b/tool/agenttool/agenttool.go
@@ -34,6 +34,7 @@ package agenttool
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/rs/xid"
@@ -153,11 +154,12 @@ func (at *AgentTool) Execute(ctx context.Context, args json.RawMessage) (json.Ra
 	// before its own tool calls.
 	ctx = agent.ContextWithConversationID(ctx, session.ConversationID(sess))
 
-	// Run agent, collecting the last assistant message as the result and the
-	// finish reason so a truncated turn can be flagged to the parent.
+	// Run agent, collecting the last assistant message, the terminal finish
+	// reason, and any fatal cause carried in an ErrorEvent.
 	var (
 		result       string
 		finishReason agent.FinishReason
+		runErr       error
 	)
 
 	for evt, err := range at.agent.Run(ctx, inv) {
@@ -169,6 +171,10 @@ func (at *AgentTool) Execute(ctx context.Context, args json.RawMessage) (json.Ra
 		case agent.MessageEvent:
 			// Capture last assistant message as result.
 			result = e.Response.Message.TextContent()
+		case agent.ErrorEvent:
+			// Non-terminal at the runtime layer: llmagent carries the fatal cause
+			// here and reports FinishReasonError on the terminal event.
+			runErr = e.Err
 		case agent.InvocationEndEvent:
 			finishReason = e.FinishReason
 		}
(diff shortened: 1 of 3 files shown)


MikolajGagatekRP:
@claude review

tool/agenttool/agenttool.go
```diff
@@ -140,26 +154,80 @@ func (at *AgentTool) Execute(ctx context.Context, args json.RawMessage) (json.Ra
...
+		// Output truncation is non-fatal: deliver the partial answer, but mark it
+		// so the parent does not mistake it for a complete result.
+		metadata = map[string]any{markerTruncated: true}
+	case agent.FinishReasonContextOverflow:
+		return nil, errors.New("agent execution failed: the conversation exceeds the model's context window")
+	case agent.FinishReasonMaxTurns:
+		return nil, errors.New("agent execution failed: maximum iterations reached")
+	case agent.FinishReasonInputRequired:
+		return nil, errors.New("agent execution failed: sub-agent requires external input, which agent-as-tool cannot provide")
+	case agent.FinishReasonInterrupted:
```
malinskibeniamin, line 207: do we want to expose this in the ADP UI inspector/playground as well?

malinskibeniamin (approved):
do we want to expose this in the ADP UI inspector/playground as well?

tool/agenttool/agenttool.go
MikolajGagatekRP: Tbh idk but imho we should

PR #188, opened 2026-07-30
feat(bedrock): give Claude models a bounded default max_tokens

MikolajGagatekRP:
## What

On the Bedrock path, `Config.MaxTokens` defaults to `nil`, and the request mapper only sends `max_tokens` when it's non-nil — so AWS applies its own default, **4096 for Claude**, which truncates ordinary agent turns (sometimes to empty output when reasoning consumes the whole budget).

This gives **Claude models on Bedrock** a bounded default output budget (16K) when the caller sets none — the same safe out-of-the-box fallback the native Anthropic provider already has (#179), so Claude behaves consistently on either path.

```go
if cfg.MaxTokens == nil && isAnthropicModel(apiModelID) {
    budget := int32(min(defaultMaxTokens, modelDef.Constraints.MaxOutputTokens))
    cfg.MaxTokens = &budget
}
```

- **Scoped to the Claude family.** Other Bedrock families (Nova, Llama, Mistral, Gemma, gpt, …) have different output caps and context windows, and their AWS defaults aren't the same footgun — a blanket default risks 400s. So the default is applied only to `anthropic.*` model IDs.
- **Clamped** to the model's `MaxOutputTokens` so it can never exceed the cap.
- **Explicit `WithMaxTokens` still wins** — the default only fills when unset. Precedence: `WithMaxTokens` > default (mirrors the anthropic provider).

## Why

Surfaced via AI-1708 (a Bedrock-Claude agent hitting the 4096 cap). #179 raised the default and made truncation non-fatal, but the default bump was anthropic-provider-only; this closes the same footgun on the Bedrock path so a Bedrock-Claude agent isn't silently left at 4096 when no harness sets a budget.

## Testing

- `TestNewModel_ClaudeDefaultMaxTokens`: a Claude model with no options → 16K, never above the output cap.
- `TestNewModel_ExplicitMaxTokensOverridesClaudeDefault`: `WithMaxTokens(4096)` wins over the default.
- `TestNewModel_NonClaudeModelHasNoDefaultMaxTokens`: a non-Claude model (Mistral) keeps `nil` — guards the Claude-only scoping.
- Built TDD (red → green). No Bedrock e2e: the send-path mechanism is already proven by the anthropic e2e; this change is the default value + scoping, fully unit-testable.

## Scope

Default value + Claude scoping only. A per-request `max_tokens` override on Bedrock (matching the anthropic `RequestOptions`) is possible future work, not included here.

### DIFF
--- a/providers/bedrock/provider.go
+++ b/providers/bedrock/provider.go
@@ -205,7 +205,6 @@ func WithCachingDisabled() ProviderOption {
 	}
 }
 
-// NewModel creates a new Bedrock model instance with the specified configuration.
 // defaultMaxTokens is a bounded fallback output budget for Claude models on
 // Bedrock. Bedrock sends no max_tokens unless one is set, so AWS applies its own
 // default — 4096 for Claude — which truncates ordinary agent turns. This gives
@@ -217,6 +216,7 @@ func WithCachingDisabled() ProviderOption {
 // conversations.
 const defaultMaxTokens = 16384
 
+// NewModel creates a new Bedrock model instance with the specified configuration.
 func (p *Provider) NewModel(modelName string, opts ...Option) (llm.Model, error) {
 	// Build the API model ID. Most Bedrock models are cross-region inference
 	// profiles, so an un-prefixed name gets the source region's geo prefix
@@ -275,7 +275,12 @@ func (p *Provider) NewModel(modelName string, opts ...Option) (llm.Model, error)
 	// Give Claude models a bounded default output budget when the caller set none,
 	// so they are not left at AWS's 4096. Scoped to Claude; clamped to the model's
 	// output cap so it can never exceed the limit. An explicit WithMaxTokens wins.
-	if cfg.MaxTokens == nil && isAnthropicModel(apiModelID) {
+	//
+	// The MaxOutputTokens > 0 guard keeps the budget >= 1 (a catalog entry with an
+	// unset cap would otherwise put max_tokens: 0 on the wire). setOptions is left
+	// untouched on purpose: this is an SDK-injected default, not a caller-set
+	// option, so it must not trip caller-facing conditional rules in Validate.
+	if cfg.MaxTokens == nil && isAnthropicModel(apiModelID) && modelDef.Constraints.MaxOutputTokens > 0 {
 		budget := int32(min(defaultMaxTokens, modelDef.Constraints.MaxOutputTokens)) //nolint:gosec // bounded: 1 <= budget <= MaxOutputTokens
 		cfg.MaxTokens = &budget
 	}
--- a/providers/bedrock/provider.go
+++ b/providers/bedrock/provider.go
@@ -206,6 +206,17 @@ func WithCachingDisabled() ProviderOption {
 }
 
 // NewModel creates a new Bedrock model instance with the specified configuration.
+// defaultMaxTokens is a bounded fallback output budget for Claude models on
+// Bedrock. Bedrock sends no max_tokens unless one is set, so AWS applies its own
+// default — 4096 for Claude — which truncates ordinary agent turns. This gives
+// Claude a safe out-of-the-box budget matching the native Anthropic provider; an
+// explicit WithMaxTokens still wins. It is scoped to Claude because the other
+// Bedrock families have different output caps and their AWS defaults are not the
+// same footgun. It is bounded rather than the model max because max_tokens
+// reserves context-window space, so defaulting to the max would 400 long
+// conversations.
+const defaultMaxTokens = 16384
+
 func (p *Provider) NewModel(modelName string, opts ...Option) (llm.Model, error) {
 	// Build the API model ID. Most Bedrock models are cross-region inference
 	// profiles, so an un-prefixed name gets the source region's geo prefix
@@ -259,6 +270,14 @@ func (p *Provider) NewModel(modelName string, opts ...Option) (llm.Model, error)
 		if err := opt(cfg); err != nil {
 			return nil, fmt.Errorf("invalid option for %s: %w", modelName, err)
 		}
+	}
+
+	// Give Claude models a bounded default output budget when the caller set none,
+	// so they are not left at AWS's 4096. Scoped to Claude; clamped to the model's
+	// output cap so it can never exceed the limit. An explicit WithMaxTokens wins.
+	if cfg.MaxTokens == nil && isAnthropicModel(apiModelID) {
+		budget := int32(min(defaultMaxTokens, modelDef.Constraints.MaxOutputTokens)) //nolint:gosec // bounded: 1 <= budget <= MaxOutputTokens
+		cfg.MaxTokens = &budget
 	}
 
 	if err := cfg.Validate(); err != nil {
--- a/providers/bedrock/default_max_tokens_integration_test.go
+++ b/providers/bedrock/default_max_tokens_integration_test.go
@@ -1,0 +1,137 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bedrock
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/redpanda-data/ai-sdk-go/llm"
+	"github.com/redpanda-data/ai-sdk-go/providers/bedrock/bedrocktest"
+)
+
+// TestBedrockOmittedMaxTokensDefault_Integration is the live boundary check that
+// establishes the fact the Claude default max_tokens injection depends on: that
+// Converse, given no maxTokens for a Claude model, applies a small implicit
+// output cap far below the model's advertised MaxOutputTokens — rather than the
+// model maximum the published InferenceConfiguration reference describes. Unit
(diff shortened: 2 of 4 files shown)


MikolajGagatekRP:
@claude review

MikolajGagatekRP:
Thanks — addressed in e3365cd:

- **Must-fix (1)** doc comment: moved the `defaultMaxTokens` const + its comment above `NewModel`'s doc comment, so `NewModel` is documented again.
- **(3)** zero-budget: added a `MaxOutputTokens > 0` guard so a catalog entry with an unset cap can't emit `max_tokens: 0`; the `nolint`'s `>= 1` invariant now holds.
- **(2)** documented why `setOptions` is intentionally left untouched (SDK-injected default, not a caller option).
- **(5)** anchored `isAnthropicModel` on the family segment: `HasPrefix("anthropic.") || Contains(".anthropic.")`.
- **Test coverage:** table-driven the default test (claude default / explicit-override-wins / non-claude-nil) and added `TestNewModel_ClaudeDefaultReachesConverseInput`, which asserts the default reaches `InferenceConfig.MaxTokens` on the Converse input — the actual bug locus.

Deferred as noted:
- **(4)** `WithThinking` interaction (`max_tokens > budget_tokens`): real but not a regression; better as a follow-up that validates `BudgetTokens < *MaxTokens` in `Validate` for a clear construction-time error.
- **(6)** `DefaultMaxTokens` field on `ModelDefinition`: worth it only if this policy recurs for other families; leaving `NewModel` as the single site for now.

MikolajGagatekRP:
@claude review

providers/bedrock/provider.go
```diff
@@ -261,6 +272,19 @@ func (p *Provider) NewModel(modelName string, opts ...Option) (llm.Model, error)
...
 
+	// Give Claude models a bounded default output budget when the caller set none,
+	// so they are not left at AWS's 4096. Scoped to Claude; clamped to the model's
+	// output cap so it can never exceed the limit. An explicit WithMaxTokens wins.
+	//
+	// The MaxOutputTokens > 0 guard keeps the budget >= 1 (a catalog entry with an
+	// unset cap would otherwise put max_tokens: 0 on the wire). setOptions is left
+	// untouched on purpose: this is an SDK-injected default, not a caller-set
+	// option, so it must not trip caller-facing conditional rules in Validate.
+	if cfg.MaxTokens == nil && isAnthropicModel(apiModelID) && modelDef.Constraints.MaxOutputTokens > 0 {
```
malinskibeniamin, line 283: [P1 Major] Verify the Bedrock default before overriding it

What: The core premise here—that omitting Converse `maxTokens` caps Claude at 4096—is not exercised across Bedrock; the added tests begin after this default is injected. The pinned AWS SDK documentation and the current [Bedrock API reference](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InferenceConfiguration.html) instead say an omitted value defaults to the model's maximum allowed value.

Why: If that published contract matches runtime for these current Claude models, this lowers the effective default from 64K/128K to 16K and causes the truncation this PR intends to prevent. This boundary claim needs real-provider evidence (`integration-tests-cross-real-boundaries`).

Suggested fix: Before merge, add or record a reproducible live Converse check for a current catalog Claude model that distinguishes omitted `maxTokens` from explicit 16K (or attach the AI-1708 request/response evidence proving the 4096 stop); remove or adjust this injection if omission is no longer 4096.

One-shot prompt: No safe one-shot—proving omitted-field behavior requires authorized Bedrock credentials and a live boundary check; unit tests cannot establish it.

MikolajGagatekRP: Good catch — agreed the premise was never exercised on the wire. The added unit tests assert config state *after* injection, so they cannot distinguish an implicit 4096-style cap from the model maximum, and the "4096 for Claude" figure traces back to a code-reading inference rather than a live observation.

I pushed a credentials-gated integration test (`TestBedrockOmittedMaxTokensDefault_Integration`) that settles it on a real boundary: it runs the same long-output prompt twice against a current catalog Claude model — once with `maxTokens` omitted (the injected default cleared so the field is truly absent) and once with it set to the injected default — and compares the observed `output_tokens`. If the omitted run truncates below the explicit run, the implicit default is the low cap the injection lifts; if it runs to the model maximum (as the InferenceConfiguration reference describes), the injection *lowers* the ceiling and the assertion fails loudly with that instruction.

Marked the PR draft until someone with authorized Bedrock creds records that run. If the result shows omission already yields the model max, I will drop the injection.

malinskibeniamin: [P1 Major] Still open: the live check times out before measuring the omitted default

What: The new boundary test never reaches either token-count assertion in credentialed CI; its first omitted `Converse` call hits the four-minute context deadline, and the required `Test` check is red.
Why: This neither verifies the claimed 4096 default nor makes the PR mergeable. The injection still contradicts the published Converse contract and may lower Claude's effective ceiling, while every credentialed `task test` now pays a four-minute failure.
Suggested fix: Replace this unbounded non-streaming 100,000-word probe with a bounded live check (or record a manual request/response) that actually distinguishes omission from 16K; keep the injection only if that evidence confirms a lower implicit cap.
One-shot prompt: In `redpanda-data/ai-sdk-go` on `bedrock-claude-default-max-tokens`, redesign `providers/bedrock/default_max_tokens_integration_test.go` so the omitted-field boundary check completes within CI and records decisive evidence, then run the targeted live test and `task test`.

CI evidence: https://github.com/redpanda-data/ai-sdk-go/actions/runs/30639790822/job/91186495533

providers/bedrock/provider.go
```diff
@@ -261,6 +272,19 @@ func (p *Provider) NewModel(modelName string, opts ...Option) (llm.Model, error)
...
+		budget := int32(min(defaultMaxTokens, modelDef.Constraints.MaxOutputTokens)) //nolint:gosec // bounded: 1 <= budget <= MaxOutputTokens
```
malinskibeniamin, line 284: [P1 Major] Make the injected default compatible with extended thinking

What: `WithThinking(16384)` leaves `MaxTokens` unset, then this line injects `maxTokens=16384` beside `budget_tokens=16384`.
Why: Bedrock requires `budget_tokens` to be strictly less than `max_tokens`, so a valid public option is accepted by `NewModel` but fails at the provider with a 400. Larger documented thinking budgets fail too. This violates `reject-invalid-explicit-input`.
Suggested fix: When the caller did not set `WithMaxTokens`, derive a thinking-aware value that is greater than `BudgetTokens` and within `MaxOutputTokens`, or reject an impossible combination during construction; add a regression for a budget at/above 16K.
One-shot prompt: In `redpanda-data/ai-sdk-go` on `bedrock-claude-default-max-tokens`, update `providers/bedrock/provider.go:283-285` so the Claude default remains strictly above an enabled thinking budget (or returns a construction error when the model cap makes that impossible), add targeted tests, then run `go test -short ./providers/bedrock`.

AWS contract: https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-extended-thinking.html

providers/bedrock/provider.go
MikolajGagatekRP: Confirmed on the wire — you were right. In a credentialed CI run against a Claude-enabled account, the omitted-`maxTokens` Converse call on `claude-sonnet-4-5` generated for the full 240s deadline instead of returning at ~4096. So Bedrock applies the model maximum when the field is absent, matching the InferenceConfiguration reference — the 4096 premise does not hold. A 16K default injection would therefore *lower* the effective output ceiling (64K→16K) for callers that omit `maxTokens`, i.e. the opposite of the intent.

Dropping the injection and closing this PR (removing the unbounded live probe with it). The truncation fix that was actually needed shipped in #179 (truncation made non-fatal + a distinct context-overflow finish reason), which is provider-agnostic and already covers the Bedrock path. Thanks for holding the line on real-boundary evidence.

providers/bedrock/provider.go
MikolajGagatekRP: With the injection dropped (closing this PR) this conflict goes away — `NewModel` no longer sets `max_tokens` next to a `WithThinking` budget, so an SDK-injected value can no longer violate the strictly-greater-than requirement. Noting for the record: if a Bedrock default is ever reintroduced, it must be thinking-aware (`max_tokens` strictly above any enabled `budget_tokens`, within the output cap) or reject the impossible combination at construction.

MikolajGagatekRP:
Closing. The boundary check this PR was gated on has now been run against a Claude-enabled account, and it refutes the premise: omitting `maxTokens` on Bedrock Converse yields the model maximum, not 4096 (matching the AWS InferenceConfiguration reference). A default injection would lower Claude's effective output ceiling rather than raise it.

The truncation-related fix that was actually needed shipped in #179 (truncation made non-fatal + a distinct context-overflow finish reason) — provider-agnostic, already covering Bedrock. Dropping this change.

MikolajGagatekRP:
Correction to my earlier note on this PR. I wrote that a credentialed run "confirmed omitted = model max" — that was based on an ambiguous timeout, not a real token count. A subsequent **bounded, credentialed measurement against raw Bedrock (no proxy)** shows the omitted-`max_tokens` default is **model-generation-dependent**:

| Model | omitted `max_tokens` result |
|---|---|
| claude-opus-5 | 4096 (cut, `max_tokens`) |
| claude-opus-4-8 | 4096 (cut, `max_tokens`) |
| claude-sonnet-5 | 4096 (cut, `length`) |
| claude-sonnet-4-6 | ~8K, ended naturally (model-max) |
| claude-sonnet-4-5 | ~23K, ended naturally (model-max) |

So the **current generation caps omitted output at exactly 4096**, while pre-Sonnet-5 models default to model-max (as the InferenceConfiguration reference describes). Both honor an explicit `max_tokens` (16384 → 16384). The default is not declared anywhere and can change.

Net for this PR: **closing stands — but the reason is the model-dependency, not "there is no 4096 cap."** A blanket provider-level default would *raise* the current-gen models (4096→16K) yet *lower* the pre-5 models (model-max→16K), so it does not belong in the SDK as an unconditional injection. If a Bedrock default is ever wanted, it must be set where the model and use-case are known (the caller setting `max_tokens` explicitly), and be model-aware + thinking-aware.

Thanks for holding the line on real-provider evidence — that is exactly what surfaced the per-generation behavior.

PR #194, opened 2026-08-12
example: add web research agent with parallel tool calls

weeco:
Interactive research assistant: decomposes a question into sub-topics, runs the searches in parallel, synthesizes a cited answer.

Covers ground the other examples don't — multiple tool calls in one turn (so they execute concurrently), a multi-turn REPL over one session, and `plugins/otel` tracing the run.

Tracing is opt-in via the standard OTel env vars only (no endpoint or auth baked in); with none set it runs on just an OpenAI + Tavily key.

Stacked on #193.


examples/web_research_agent/tools.go
```diff
@@ -0,0 +1,225 @@
...
+
+// ---- fetch_url (real HTTP GET) --------------------------------------------
+
+// FetchURLTool fetches a URL and returns its (readable) content.
+type FetchURLTool struct {
+	client *http.Client
+}
+
+func NewFetchURLTool() *FetchURLTool {
+	return &FetchURLTool{client: &http.Client{Timeout: 20 * time.Second}}
```
malinskibeniamin, line 143: [P0 Blocker] Do not expose the host network through a model-controlled client

What: `fetch_url` accepts any HTTP(S) URL and uses the default transport, including its redirect handling. A malicious search result or fetched page can prompt the model to request loopback, RFC1918, or link-local metadata URLs; the response is then returned to the model and, when tracing is on, recorded in spans. An overlay test against this code confirmed that `127.0.0.1` is fetched and its body returned.

Why: Running the research example on a developer or cloud host exposes internal services and credentials to indirect prompt injection. The README warning cannot make this safe because consuming untrusted web content is the example's normal flow.

Suggested fix: Reuse `tool/builtin/webfetch`, which already blocks private/reserved IPs, prevents DNS-rebinding and redirect bypasses, restricts content types, and fences untrusted content; or apply the same dial and redirect checks here. This is the `tenant-egress-routes-through-safedial` rule.

Verify: `cd examples/web_research_agent && go test -run 'TestFetchURLRejectsPrivateTargets|TestFetchURLRejectsPrivateRedirect' ./...`

simon0191 (approved):

examples/web_research_agent/tools.go
weeco: Follow-up in #198: swapped the hand-rolled fetch_url for tool/builtin/webfetch as suggested; web_search stays custom as the tool-authoring demo.

PR #195, opened 2026-08-12
chore(openai): remove retired gpt-5.2-chat-latest and gpt-5.3-chat-latest

weeco:
OpenAI pulled both models — the API now returns `Model not found`, failing `TestOpenAIConformance_Integration/TestAllSupportedModels` and `TestAllModelsReasoningEffortsIntegration` on every PR. `main` was green at de1a67d earlier the same day, so they went away today.

Removes the catalog entries, the `ModelGPT5_2Instant` / `ModelGPT5_3ChatLatest` constants, and the test cases pinning their reasoning-effort sets — same approach as fd3cac7 for claude-opus-4-1.

Needs to land before #193 / #194 can go green.

### DIFF
--- a/providers/openai/models.go
+++ b/providers/openai/models.go
@@ -198,30 +198,6 @@ var supportedModels = map[string]ModelDefinition{
 		SupportedReasoningEfforts: []ReasoningEffort{ReasoningEffortNone, ReasoningEffortLow, ReasoningEffortMedium, ReasoningEffortHigh, ReasoningEffortXHigh},
 		Pricing:                   pricing.FlatInfo(0.875, 7.00, 0.175),
 	},
-	ModelGPT5_2Instant: {
-		Name:  ModelGPT5_2Instant,
-		Label: "OpenAI GPT-5.2 Instant",
-		Capabilities: llm.ModelCapabilities{
-			Streaming:        true,
-			Tools:            true,
-			JSONMode:         true,
-			StructuredOutput: true,
-			Vision:           true,
-			Audio:            true,
-			MultiTurn:        true,
-			SystemPrompts:    true,
-			Reasoning:        true,
-		},
-		Constraints: llm.ModelConstraints{
-			TemperatureRange:  [2]float64{0.0, 2.0},
-			MaxInputTokens:    400000, // 400K context window
-			MaxOutputTokens:   128000, // 128K output tokens
-			SupportedParams:   []string{"temperature", "top_p", "max_tokens", "frequency_penalty", "presence_penalty", "seed", "reasoning_effort", "reasoning_summary"},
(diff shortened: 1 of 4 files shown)


malinskibeniamin (commented):

providers/openai/models.go
```diff
@@ -247,32 +223,6 @@ var supportedModels = map[string]ModelDefinition{
 		Pricing:                   pricing.FlatInfo(10.50, 84.00, 0),
 	},
 
-	// GPT-5.3 Series
-	ModelGPT5_3ChatLatest: {
```
malinskibeniamin, line 251: **Priority: P1**

**Retired model IDs are still accepted, now with the wrong family's constraints and pricing.**

Removing the `ModelGPT5_3ChatLatest` and `ModelGPT5_2Instant` (line 201) catalog entries does not reject the retired IDs, because `resolveModelFamily` (`providers/openai/models.go:49`) falls back to longest-prefix matching over `supportedModels` keys:

- `"gpt-5.3-chat-latest"` has prefix `"gpt-5"` → resolves to the `gpt-5` family
- `"gpt-5.2-chat-latest"` has prefix `"gpt-5.2"` → resolves to the `gpt-5.2` family

`Provider.NewModel` (`providers/openai/provider.go:137-144`) therefore finds a definition and succeeds. It then calls `NewCompatModel`, which sends `modelName` verbatim as the API model ID while applying the *wrong* family's `Constraints`, `SupportedReasoningEfforts`, and `Pricing`. Consequences for a caller still passing a retired ID: option validation now accepts reasoning efforts the retired model never supported (e.g. `none`/`minimal` via the `gpt-5` family, where previously only `medium` was allowed), cost accounting is computed from the wrong price table (`gpt-5.3-chat-latest` was 1.75/14.00/0.175, `gpt-5` is not), and the only failure signal is a late remote API error instead of the local `unsupported OpenAI model` this chore implies.

Correction: gate prefix fallback so it cannot absorb non-snapshot suffixes — either keep an explicit retired/denied-ID set checked before resolution, or require the remainder after the family prefix to match a snapshot pattern (`-YYYY-MM-DD`) rather than any `-…` suffix. Add table cases asserting `NewModel("gpt-5.2-chat-latest")` and `NewModel("gpt-5.3-chat-latest")` both return `unsupported OpenAI model`, alongside the existing `TestNewModelWithTimestampedName` coverage.

Verify: `go test ./providers/openai/ -run 'TestProviderModels|TestNewModelWithTimestampedName|TestReasoningEffort'`

*Automated `/review`.*

malinskibeniamin (approved):
LGTM but I think we need to also ensure aliases like `gpt-5.3` are removed/retired as well

simon0191 (approved):

PR #199, opened 2026-08-19
AI-1847: Model catalog rework: shared catalog package, lifecycle metadata, breaking discovery API

weeco:
Tracks [AI-1847](https://redpandadata.atlassian.net/browse/AI-1847) (Model Catalog Improvements).

## Why

The immediate driver is the AI Gateway: its console needs to show a real model picker, with cost class, retirement dates, and which model in a family is the latest. The old surfaces carried none of that. Beyond the missing data, the old storage kept hurting:

- One model had to be written out many times. Opus 4.8 meant one full definition for Anthropic and four more for Bedrock (us, eu, jp, global), each repeating data that never varies per host, such as the display name, release date, and knowledge cutoff. Vertex AI would have added more copies.
- Those hand-written copies disagreed. This happened twice in this repo.
- When a model retired we deleted its entry, so historical usage could no longer be priced and old failures could no longer be explained.
- Questions like "is this the current generation" or "what replaced it" had no API, and where an answer was stored it went stale.

## What this is

One shared, validated `catalog` package replaces the four per-provider `ModelDefinition` structs. Every offering carries capabilities, constraints, modalities, reasoning controls, pricing, and lifecycle, and the questions a console needs answered (current generation, successor, retired, price tier) are computed from that data at read time.

**Breaking:** `llm.ModelDiscoveryInfo`, every `Models()`, and every `ModelPricing()` are removed. `provider.Catalog()` is the single discovery and pricing surface. The old to new mapping is in **MIGRATION.md**. The sequence ships in one release, with no tag in between.

## The concepts

**Facts vs entries.** Every piece of metadata has exactly one home. Truths that hold everywhere a model is served (display name, series, release date, knowledge cutoff) are facts, written once in `catalog/facts_data.go` and keyed by a canonical ModelID such as `anthropic/claude-opus-5`. Everything that depends on the host (capabilities, token limits, pricing, lifecycle) lives on the entry, one per invokable ID per provider, authored in `providers/*/models.go`. Entries reference facts, and `catalog.New` fails on an unregistered ModelID, so two providers cannot hold different copies of the same model's facts.

**Derive, never store.** Anything computable from the rest of the catalog is computed at read time: generation order from release dates within a series, successors from series order (a vendor-announced `ReplacedBy` wins), retirement from lifecycle dates, price tier from the rate card. Adding a model never edits an existing entry, and derived answers cannot go stale. Everything clock-dependent lives on `Catalog.Now()` / `.At()` views, so anything built from the catalog alone is reproducible.

**Append-only.** Retired offerings keep their entries, with `Retires` in the past and `Deprecated:` markers on their constants, so historical usage stays priceable and old failures stay explainable. Three models the old catalogs had deleted on retirement were restored this way.

**Unknown means stop enforcing.** When `Resolve` does not recognize a model, callers must not guess capabilities or constraints for it, and pricing reports it as unpriced, never as free.

**The snapshot.** `catalog/snapshot.json` is the whole expanded catalog (51 models, 94 offerings), committed and CI-checked. Its diff is the review surface for any catalog change, and it is the read format for non-Go consumers such as the gateway console.

**Bedrock families.** One `family` declaration per model expands into its geo-profile variants (us, eu, jp, au, global): 48 offerings from 18 declarations. Each profile variant carries an `inference_geo` attribute, so geography filters read structured data instead of parsing ID prefixes.

## Behavior changes

- Stricter OpenAI reasoning-effort validation: a requested effort against a model with no declared efforts is now rejected.
- Only version stamps prefix-resolve. A suffix maps onto its family only when it is a date or revision stamp (`-20250929`, `-2025-04-16`, `-001`, `@001`). Short version bumps such as `gpt-5.7` or `claude-opus-5-1` are unknown models and `NewModel` rejects them; the previous resolver silently mapped them onto `gpt-5` / `claude-opus-5` with that model's constraints and pricing.
- Alias and snapshot resolution improved on OpenAI and Google.
- `llm.Response.InvokedModelID` reports catalog offering IDs where the provider reports a snapshot the catalog recognizes.

## Verification

All unit tests green, `task lint:new` 0 issues, license check, examples build. `TestCommittedSnapshotIsFresh` in `cmd/catalog-snapshot` fails the ordinary unit-test job when `catalog/snapshot.json` is stale, so no workflow needs a dedicated `task catalog:check` step. Full `task lint` fails identically on `main`. Integration tests need API keys and were not run here; one keyed run before merge would cover the conformance changes.

## Gateway impact: supersedes RFC-0010 §2.2

RFC-0010 (Cost Reporting) builds the gateway's pricing catalog from `openai.ModelPricing()`, `anthropic.ModelPricing()`, `google.ModelPricing()` and `bedrock.ModelPricing()`. Those functions are removed here, so that section of the RFC is superseded by `provider.Catalog().PricingByID()`; the mapping is in **MIGRATION.md**. This is not only documentation: `apps/aigw` calls the removed surfaces today and will need the swap when it bumps `ai-sdk-go` past this release.

- `internal/spending/pipeline.go` and `internal/spending/mockfixtures/realistic.go`: `ModelPricing()` → `Catalog().PricingByID()`. Billing lookups for timestamped `InvokedModelID`s go through `Catalog().Resolve` first (see MIGRATION.md).
- `internal/models/registry.go`: the `modelProvider` interface (`Models() []llm.ModelDiscoveryInfo`) → `catalog.Provider` (`Catalog() *catalog.Catalog`). `Catalog().All()` carries display name, capabilities, constraints, and `Reasoning.Efforts` directly, so the per-model `NewModel` probe for supported efforts and the hand-patched `gpt-5.6` alias row both go away: aliases are catalog data. `Catalog().Now()` gives the lifecycle views the picker needs.

## Follow-ups (out of scope, recorded in the design doc)

Weekly lifecycle-sync automation, per-offering tuning defaults and a tool-output cap, a remote-refreshable catalog, price-tier threshold sign-off with the gateway team.

Orientation guide: https://claude.ai/code/artifact/ee37a94d-1f52-4a18-b168-ae4334d882d2

[AI-1847]: https://redpandadata.atlassian.net/browse/AI-1847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MaciejWrzesniewski-Redpanda (commented):
### The removed `bedrock.ThinkingSupport` type is still in the tree

MIGRATION.md line 55 lists `bedrock.ThinkingSupport` as removed, mapped to `catalog.ReasoningSupport`. The type is still declared at `providers/bedrock/models.go:225`. I grepped the repo and nothing constructs or references it - the only other mention is a comment in `catalog/reasoning.go:20` that calls it "the former bedrock.ThinkingSupport". Its own doc comment is stale too - it talks about `supportedModels` entries and `NewModel` validating against it, both of which this PR removed. The `unused` linter does not flag exported symbols, so CI stays green over it. I would just delete the type so the public surface matches the migration guide.

---

### Mind linking this to AI-1847?

I could not see the ticket referenced anywhere on the PR - branch, title, commits, or body. This is the model catalog rework that AI-1847 (Model Catalog Improvements) tracks, so dropping the key into the title or a commit would let the ticket track its own work, and let us close AI-1904 as done by this.

---

### This supersedes RFC-0010's pricing-catalog wiring

RFC-0010 (Cost Reporting, `apps/aigw/docs/rfcs/0010-cost-reporting.md`) section 2.1 builds the gateway's pricing catalog from `openai.ModelPricing()`, `anthropic.ModelPricing()` and the rest, passed into `pricing.NewCatalog(...)`. This PR removes all of those in favour of `provider.Catalog().PricingByID()`, so that code no longer compiles. I think the PR description should say it supersedes RFC-0010 section 2.1, and point at MIGRATION.md for the swap, so the gateway folks see it coming.

MaciejWrzesniewski-Redpanda (approved):

PR #200, opened 2026-08-19
deps: bump openai-go to v3.43.0 to fix SSE keepalive streaming

birdayz:
## What

Bump `github.com/openai/openai-go/v3` from v3.42.0 to v3.43.0, plus a regression test in `providers/openaicompat` that pins the behaviour.

## Why

openai-go v3.42.0 breaks every streaming call against a provider that emits SSE comment keepalives. OpenRouter sends them while it queues a request, and some vLLM and LiteLLM deployments do the same:

```
: OPENROUTER PROCESSING
<blank>
```

`eventStreamDecoder.Next()` skips the comment line, then the blank line that follows it hits the "dispatch event on an empty line" branch and dispatches an event with an empty data buffer. `Stream[T].Next()` calls `json.Unmarshal([]byte{}, &chunk)`, which fails and kills the stream. `openaicompat` surfaces that as

```
API call failed: unexpected end of JSON input
```

before a single chunk reaches the caller. It is intermittent, which is what made it look like a misconfiguration: keepalives only appear when the upstream holds the connection open — queuing, cold start, reasoning-heavy models. Fast responses stream fine, so connection checks and non-streaming calls all pass.

Hit in production on a customer running OpenRouter through `openaicompat`.

## Implementation details

v3.43.0 adds the guard the WHATWG SSE spec requires — do not dispatch when the data buffer is empty. That is the whole release, four lines in `packages/ssestream/ssestream.go`:

```go
if len(txt) == 0 {
    if data.Len() == 0 {
        event = ""
        continue
    }
    ...
```

The regression test streams the customer's byte sequence through an `httptest` server with keepalives interleaved between real chunks. It fails against v3.42.0 with the exact production error and passes from v3.43.0 on, so a downgrade cannot slip back in unnoticed.

v3.43.0 rather than the current v3.52.0: one patch release, one four-line fix, nothing else to review. v3.52.0 is ten minor versions of API drift for no benefit here.

Consumers who require openai-go directly are already unblocked by MVS picking the maximum; this bump is so a fresh consumer of the SDK does not inherit the broken version.

## References

- AI-1902
- openai-go v3.43.0: https://github.com/openai/openai-go/releases/tag/v3.43.0


birdayz:
@claude review this PR

splittingfield (approved):

PR #201, opened 2026-08-19
fix(openaicompat): classify decode and transport failures

birdayz:
## What

`classifyError` in `providers/openaicompat` now produces a structured `*llm.ProviderError` for failures that never carried an HTTP status: JSON decode failures, transport failures, and errors the upstream delivered inside an SSE data frame.

Follow-up to the openai-go v3.43.0 bump (#200), independent of it.

## Why

`classifyError` only built a `ProviderError` for `*openai.Error`. Everything else was returned raw, so any failure below HTTP reached the caller with no category, no code and no status.

That is how an openai-go SSE decoding bug (AI-1902) presented in production. Streaming against OpenRouter died with

```
agent: model generation failed: API call failed: unexpected end of JSON input
```

category `unknown`, code `-1`. Nothing in the error said the failure was a decode failure rather than something the upstream rejected, and nothing said a retry was worth attempting. Diagnosing it took a raw SSE capture off the customer's gateway. The same error now reads

```
API call failed: server error: [response_decode_error] could not decode provider response: unexpected end of JSON input
```

## Implementation details

| Failure | Base | Code | Retryable |
| --- | --- | --- | --- |
| `*json.SyntaxError`, `*json.UnmarshalTypeError` | `ErrServerError` | `response_decode_error` | yes |
| `net.Error`, `io.ErrUnexpectedEOF`, `io.EOF` | `ErrServerError` | `transport_error` | yes |
| `*ssestream.StreamError` | `ErrAPICall` | `stream_error` | no |

**Why decode failures are retryable.** The usual cause is protocol noise on a slow upstream response — keepalives, a truncated frame — and the same request often survives a second attempt. AI-1902 was intermittent for exactly this reason: it only fired when the upstream held the connection open.

**Why `stream_error` is not.** The payload is opaque at this layer — a mid-stream error object could be a rate limit or a hard model failure, and there is no status code to tell them apart. A mid-stream failure is as likely to be terminal as transient, so it does not claim to be retryable.

**Context errors are returned untouched.** `ProviderError.Unwrap` yields only `Base`, so wrapping a `context.Canceled` would sever the chain and break `errors.Is(err, context.Canceled)` at the call site. Checked before everything else, since a `context.DeadlineExceeded` inside a `*url.Error` also satisfies `net.Error`.

Unrecognised errors still pass through unchanged, so nothing that already worked changes shape.

Codes are unexported. The goal is a log line an operator can read, not new public API surface.

**Scope.** `providers/openai`, `anthropic`, `google` and `bedrock` have the same gap in their own `classifyError`. Left alone here to keep this reviewable; folding the non-HTTP half into a shared helper is the obvious follow-up if the team wants it.

Tests cover each branch plus the end-to-end path — a malformed SSE frame from an `httptest` upstream, through `GenerateEvents`, asserting the category, code, retryability and the exact message the caller sees.

## References

- AI-1902
- #200 (the openai-go bump that fixes the bug this made undiagnosable)

### DIFF
--- a/providers/openaicompat/errors.go
+++ b/providers/openaicompat/errors.go
@@ -15,11 +15,36 @@
 package openaicompat
 
 import (
+	"context"
+	"encoding/json"
 	"errors"
+	"fmt"
+	"io"
+	"net"
 
 	"github.com/openai/openai-go/v3"
+	"github.com/openai/openai-go/v3/packages/ssestream"
 
 	"github.com/redpanda-data/ai-sdk-go/llm"
+)
+
+// Codes for failures that never got far enough to carry an HTTP status. An
+// OpenAI-compatible upstream has no error taxonomy of its own for these, so we
+// supply one: without it the errors below reach the caller with no category and
+// no code, which is exactly how an openai-go SSE decoding bug presented as a
+// bare "unexpected end of JSON input" with nothing in the error to point at.
+const (
+	// codeResponseDecode: the upstream returned bytes we could not decode as an
+	// OpenAI-shaped response.
+	codeResponseDecode = "response_decode_error"
+
+	// codeTransport: the request failed below HTTP — connection reset,
+	// truncated body, DNS failure.
+	codeTransport = "transport_error"
+
+	// codeStreamError: the upstream reported a failure inside an SSE data frame
+	// instead of as an HTTP error.
+	codeStreamError = "stream_error"
 )
 
 // classifyError maps OpenAI SDK errors to *llm.ProviderError with the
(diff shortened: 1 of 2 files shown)


birdayz:
@claude review this PR

splittingfield (approved):

birdayz:
## Codex review (post-merge, archived for the follow-up)

This PR was merged at 22:29 UTC, roughly a minute after the review round was
kicked off. Both reviewers finished afterwards. The Claude review is the comment
above; the independent codex (gpt-5.5 via Bedrock) review never got posted
because the PR was already merged by then, so it is recorded here rather than
lost.

Findings: P1=0, P2=1, P3=0.

**[P2] `net.Error` makes permanent failures retryable** — `providers/openaicompat/errors.go:106`

`*url.Error` itself satisfies `net.Error`, so the transport branch is
effectively "anything `http.Client` handed back". A permanent client-side
failure — malformed base URL producing `unsupported protocol scheme`, or a TLS
certificate verification failure — becomes a retryable `ErrServerError` and
burns the full retry budget on something that cannot succeed. Suggestion: check
`netErr.Timeout()`/`Temporary()` or a known set of network/EOF causes before
setting `Retryable`.

The Claude review reached the same place from a different angle (the note about
`providers/google/errors.go:38-47` classifying `net.Error` only when `Timeout()`
is true), so the two reviewers agree this predicate is the weakest part of the
change.

CI note: the `Test` failure on this branch and on the merge commit was
`TestRegistry_WebfetchToolWithLLM_Integration` — a live-LLM assertion on model
prose ("Got: Done."), unrelated to this diff. It passed on re-run, and main is
green at `9d9399e3`.

PR #202, opened 2026-08-19
fix(openaicompat): retry transport failures instead of guessing at them

birdayz:
## What

`openaicompat` classifies any non-HTTP failure it does not specifically recognise as a retryable `transport_error`, rather than testing which transport failures are transient.

`*json.UnmarshalTypeError` no longer retries; `*json.SyntaxError` still does.

## Why

**`net.Error` does not describe the set of transport failures.** An HTTP/2 stream reset arrives as net/http's `http2StreamError`: unexported, not a `net.Error`, and unreachable by `errors.As` because net/http vendors its own copy of x/net/http2. Any entry condition narrower than "everything else" misses it, and it then leaves the classifier with no code, no category and `Retryable` unset, which `llm.IsRetryable` reads as false. h2 is the default for `https://` upstreams and OpenAI/OpenRouter-class gateways emit server-initiated resets routinely, so this is the most common transient failure on the default protocol.

**The set cannot be enumerated.** Socket errnos differ per platform — Winsock's `WSAECONNREFUSED` is not `syscall.ECONNREFUSED`, so an errno table classifies every socket failure permanent on Windows. Proxies invent their own hangups. New protocols add more. Under an allowlist every condition not yet enumerated becomes permanent, which is the wrong way to be wrong.

**The two directions cost differently.** Retrying a permanent failure — a scheme typo, an untrusted certificate — costs three attempts at 200/400/800ms (`plugins/retry` defaults) and then reports the error it would have reported anyway. Refusing to retry a transient one fails the caller's whole task. Roughly 1.4 seconds against a dead task is not a trade worth a taxonomy to win.

`*json.UnmarshalTypeError` is the one case where retrying is provably pointless and detection is free: well-formed JSON whose shape disagrees with our structs replays identically. It becomes a non-retryable `ErrResponseMapping`. `*json.SyntaxError` is what a truncated or interleaved stream produces, so it stays retryable.

## Implementation details

`classifyNonHTTPError` keeps its recognised shapes and defaults the rest:

| failure | Base | Code | Retryable |
| --- | --- | --- | --- |
| `context.Canceled` / `DeadlineExceeded` | returned untouched | — | — |
| error inside an SSE frame | `llm.ErrAPICall` | `stream_error` | no |
| `*json.UnmarshalTypeError` | `llm.ErrResponseMapping` | `response_decode_error` | no |
| `*json.SyntaxError` | `llm.ErrServerError` | `response_decode_error` | yes |
| anything else | `llm.ErrServerError` | `transport_error` | yes |

Context errors are returned unwrapped because `ProviderError.Unwrap` yields only `Base`, so classifying them would sever the chain and break `errors.Is(err, context.Canceled)` at the call site. A `WithTimeout` expiry surfaces as `context.DeadlineExceeded` and is therefore also passed through.

Nothing in the transport arm inspects the error's shape, so there are no per-platform files and no build tags. `go vet` is clean across linux, darwin, windows, freebsd, wasip1, plan9 and js.

`TestHTTP2StreamResetIsRetryable` pins the arm that has no other observable contract, driving a real reset through the provider against an h2 `httptest` server. It aborts the response with `panic(http.ErrAbortHandler)`: a short `Content-Length` yields a plain EOF and does not reach this path.

## Not addressed

- `ProviderError.Unwrap` returns only `Base`, so classifying severs the original cause — `errors.Is(err, syscall.ECONNREFUSED)` on the result is false. Fixing it means giving `llm.ProviderError` a wrapped cause and a multi-error `Unwrap`, which touches every provider.
- `insufficient_quota` arrives as HTTP 429 and is classified by status, so a permanent billing failure is retried like a rate limit. Belongs in a change to the HTTP path.
- openai-go's client retries transport errors twice before `classifyError` sees anything, so two retry loops multiply. `option.WithMaxRetries(0)` and letting the retry interceptor own the policy is the fix.
- A provider-side timeout is indistinguishable from a caller deadline by error value, so it is not marked retryable. `model.go:74` and `:217` both hold the `ctx`, and `ctx.Err() == nil` beside a deadline error would identify it — at the cost of threading `ctx` through the exported `ResponseMapper.FromProviderError`.
- `providers/openai`, `anthropic`, `google` and `bedrock` return non-HTTP errors raw.

## References

- AI-1902
- #200, #201

### DIFF
--- a/providers/openaicompat/errors.go
+++ b/providers/openaicompat/errors.go
@@ -19,8 +19,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
-	"net"
 
 	"github.com/openai/openai-go/v3"
 	"github.com/openai/openai-go/v3/packages/ssestream"
@@ -28,28 +26,13 @@ import (
 	"github.com/redpanda-data/ai-sdk-go/llm"
 )
 
-// Codes for failures that never got far enough to carry an HTTP status. An
-// OpenAI-compatible upstream has no error taxonomy of its own for these, so we
-// supply one: without it the errors below reach the caller with no category and
-// no code, which is exactly how an openai-go SSE decoding bug presented as a
-// bare "unexpected end of JSON input" with nothing in the error to point at.
 const (
-	// codeResponseDecode: the upstream returned bytes we could not decode as an
-	// OpenAI-shaped response.
 	codeResponseDecode = "response_decode_error"
-
-	// codeTransport: the request failed below HTTP — connection reset,
-	// truncated body, DNS failure.
-	codeTransport = "transport_error"
-
-	// codeStreamError: the upstream reported a failure inside an SSE data frame
-	// instead of as an HTTP error.
-	codeStreamError = "stream_error"
+	codeTransport      = "transport_error"
+	codeStreamError    = "stream_error"
 )
 
-// classifyError maps OpenAI SDK errors to *llm.ProviderError with the
-// appropriate sentinel base and Retryable flag.
-// This is used by both the model and response mapper.
+// classifyError maps SDK errors to *llm.ProviderError.
 func classifyError(err error) error {
 	if err == nil {
 		return nil
@@ -63,22 +46,17 @@ func classifyError(err error) error {
 	return classifyNonHTTPError(err)
 }
 
-// classifyNonHTTPError handles the failures that carry no HTTP status: decode
-// failures, transport failures, and errors the upstream delivered inside an SSE
-// frame. Anything it does not recognise is returned unchanged.
 func classifyNonHTTPError(err error) error {
-	// Cancellation and deadlines belong to the caller, not to the provider.
-	// Return them untouched: ProviderError.Unwrap yields only Base, so wrapping
-	// would sever the chain and break errors.Is at the call site.
+	// Returned untouched: ProviderError.Unwrap yields only Base, so wrapping
+	// these would break errors.Is against them at the call site.
 	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 		return err
 	}
 
+	// Not retryable: the frame's payload is opaque here, and a mid-stream
+	// failure is as likely to be terminal as transient.
 	var streamErr *ssestream.StreamError
 	if errors.As(err, &streamErr) {
-		// No status code to map — the error arrived as a data frame. Not marked
-		// retryable: the payload is opaque at this layer, and a mid-stream
-		// failure is as likely to be terminal as transient.
 		return &llm.ProviderError{
 			Base:    llm.ErrAPICall,
 			Code:    codeStreamError,
@@ -86,14 +64,19 @@ func classifyNonHTTPError(err error) error {
 		}
 	}
 
-	var (
-		syntaxErr *json.SyntaxError
-		typeErr   *json.UnmarshalTypeError
-	)
+	// Well-formed JSON disagreeing with our structs replays identically.
+	var typeErr *json.UnmarshalTypeError
+	if errors.As(err, &typeErr) {
+		return &llm.ProviderError{
+			Base:    llm.ErrResponseMapping,
+			Code:    codeResponseDecode,
+			Message: fmt.Sprintf("could not decode provider response: %s", err),
+		}
+	}
 
-	if errors.As(err, &syntaxErr) || errors.As(err, &typeErr) {
-		// Retryable: the usual cause is protocol noise on a slow upstream
-		// response, which the same request often survives on a second attempt.
+	// A mangled stream; a second attempt can land on a clean response.
+	var syntaxErr *json.SyntaxError
+	if errors.As(err, &syntaxErr) {
 		return &llm.ProviderError{
 			Base:      llm.ErrServerError,
 			Code:      codeResponseDecode,
@@ -102,17 +85,17 @@ func classifyNonHTTPError(err error) error {
 		}
 	}
 
-	var netErr net.Error
-	if errors.As(err, &netErr) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, io.EOF) {
-		return &llm.ProviderError{
-			Base:      llm.ErrServerError,
-			Code:      codeTransport,
-			Message:   fmt.Sprintf("transport failure: %s", err),
-			Retryable: true,
-		}
+	// Do not narrow this to net.Error or an errno set. An h2 stream reset is
+	// net/http's unexported http2StreamError: not a net.Error, unreachable by
+	// errors.As, and h2 is the default for https upstreams. Whatever such a
+	// check missed would silently become permanent, which costs the caller's
+	// whole task where a needless retry costs three short attempts.
+	return &llm.ProviderError{
+		Base:      llm.ErrServerError,
+		Code:      codeTransport,
+		Message:   fmt.Sprintf("transport failure: %s", err),
+		Retryable: true,
 	}
-
-	return err
 }
 
 // classifyHTTPError maps an OpenAI HTTP API error to a *llm.ProviderError.
--- a/providers/openaicompat/errors_test.go
+++ b/providers/openaicompat/errors_test.go
@@ -40,12 +40,17 @@ func TestClassifyError_Nil(t *testing.T) {
 	assert.NoError(t, classifyError(nil))
 }
 
-func TestClassifyError_UnknownError(t *testing.T) {
+func TestClassifyError_UnrecognisedIsRetryableTransport(t *testing.T) {
 	t.Parallel()
 
-	err := errors.New("something unexpected")
-	result := classifyError(err)
-	assert.Equal(t, err, result)
+	result := classifyError(errors.New("something unexpected"))
+
+	var pe *llm.ProviderError
+	require.ErrorAs(t, result, &pe)
+	require.ErrorIs(t, pe, llm.ErrServerError)
+	assert.Equal(t, codeTransport, pe.Code)
+	assert.True(t, pe.Retryable)
+	assert.True(t, llm.IsRetryable(result))
 }
 
 func TestClassifyHTTPError(t *testing.T) {
@@ -127,7 +132,7 @@ func TestClassifyError_DecodeFailure(t *testing.T) {
 	assert.Contains(t, pe.Message, "unexpected end of JSON input")
 }
 
-func TestClassifyError_DecodeTypeMismatch(t *testing.T) {
+func TestClassifyError_DecodeTypeMismatchIsNotRetryable(t *testing.T) {
 	t.Parallel()
 
 	var target struct {
@@ -141,9 +146,10 @@ func TestClassifyError_DecodeTypeMismatch(t *testing.T) {
 
 	var pe *llm.ProviderError
 	require.ErrorAs(t, result, &pe)
-	require.ErrorIs(t, pe, llm.ErrServerError)
+	require.ErrorIs(t, pe, llm.ErrResponseMapping)
 	assert.Equal(t, codeResponseDecode, pe.Code)
-	assert.True(t, pe.Retryable)
+	assert.False(t, pe.Retryable, "a type mismatch replays identically")
+	assert.False(t, llm.IsRetryable(result))
 }
 
 func TestClassifyError_StreamError(t *testing.T) {
@@ -252,4 +258,47 @@ func TestStreamDecodeErrorIsClassified(t *testing.T) {
 		"API call failed: server error: [response_decode_error] "+
 			"could not decode provider response: unexpected end of JSON input",
 		streamErr.Error())
+}
+
+// Guards the classifier's transport default against being narrowed; see the
+// comment on that arm in errors.go.
+func TestHTTP2StreamResetIsRetryable(t *testing.T) {
+	t.Parallel()
+
+	server := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"id":"gen-1","object":"chat.completion","choices":[`))
+
+		if f, ok := w.(http.Flusher); ok {
+			f.Flush()
+		}
+
+		// A short Content-Length yields a plain EOF and misses this path.
+		panic(http.ErrAbortHandler) //nolint:forbidigo // aborts h2 with RST_STREAM
+	}))
+	server.EnableHTTP2 = true
+	server.StartTLS()
+	t.Cleanup(server.Close)
+
+	provider, err := NewProvider("sk-test-key",
+		WithBaseURL(server.URL+"/v1"), WithHTTPClient(server.Client()))
+	require.NoError(t, err)
+
+	model, err := provider.NewModel("m")
+	require.NoError(t, err)
+
+	_, err = model.Generate(t.Context(), &llm.Request{
+		Messages: []llm.Message{{
+			Role:    llm.RoleUser,
+			Content: []llm.Part{llm.NewTextPart("hi")},
+		}},
+	})
+	require.Error(t, err)
+	require.ErrorContains(t, err, "INTERNAL_ERROR", "expected an h2 stream reset")
+
+	var pe *llm.ProviderError
+	require.ErrorAs(t, err, &pe)
+	assert.Equal(t, codeTransport, pe.Code)
+	assert.True(t, pe.Retryable)
+	assert.True(t, llm.IsRetryable(err))
 }


birdayz:
@claude review this PR

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=3 · P3=0

The new classifiers miss retryable transport failures in common EOF and Windows socket cases, and they retry some permanent numeric 4xx stream errors.

Full review comments:

- [P2] Keep plain EOF transport failures retryable — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors.go:177-177
  When a server accepts a connection and closes it before sending a response, `net/http` reports a `*url.Error` wrapping plain `io.EOF`, not `io.ErrUnexpectedEOF`. This still enters the transport branch via `net.Error`, but `transportRetryable` now returns false, so a dropped connection is surfaced as a permanent API-call failure instead of using the retry path that handled `io.EOF` before.

- [P2] Treat numeric 4xx stream codes as permanent — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors.go:274-278
  When a gateway sends a numeric stream error code outside the few cases above, such as `402`, `404`, or `422`, `rawCodeString` converts it to a string and this default classifies it as retryable `ErrServerError`. Since these are HTTP status codes from the stream payload, numeric 4xx values should be classified like `classifyStatusCode` rather than burning the retry budget on permanent request/account errors.

- [P2] Include WinSock errnos in retry classification — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors.go:138-145
  On Windows, real socket failures wrap WinSock errno values such as `syscall.WSAECONNRESET`/`WSAECONNABORTED` (and connection refused as WSA 10061), not these invented POSIX `syscall.ECONN*` constants. Those errors still reach this classifier through `*url.Error`, but the errno lookup falls through and marks reset/refused connections non-retryable on Windows.

Script done on 2026-08-20 00:58:20+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 1 summary (commits 13ad6e6a..ce7ff09f)

Two reviewers, one shared headline finding: the allowlist was missing conditions that are unambiguously transient — the same misclassification this branch exists to fix, pointing the other way. That was worth catching. Every code finding is fixed; the three design points are answered below with reasons rather than changes.

### All items addressed

| Round | Item: Fix | Commit |
|-------|-----------|--------|
| 1 | Bare `io.EOF` under a transport wrapper judged permanent: added to `transportRetryable`, not to the entry condition — a clean end of stream cannot reach the function, which is what makes the arm safe. Pinned by a unit test and by an end-to-end case driving a server that reads the request and hangs up | requested by @claude and codex [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Winsock errnos never matched, so every dropped connection on Windows was permanent: table now carries the Winsock analogues alongside the POSIX names. Exercised on every platform, since the lookup is numeric | requested by @claude and codex [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Numeric SSE error codes hand-enumerated with a retryable default, so a permanent 402/404 got hammered: seven string arms replaced by `strconv.Atoi` + `classifyStatusCode`. Picks up 529 for free, one definition of what a status means | requested by @claude and codex [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Preferring `type` over `code` lost OpenAI's own 429 shape (`type: "requests"`, `code: "rate_limit_exceeded"`): now takes whichever of the two the taxonomy recognises and reports that one as `Code`, so `type`-only frames keep working | requested by @claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Codeless frame reused `unparseable_stream_error` and diverged from anthropic's retryability while the comment claimed parity: separate `unknown_stream_error` constant added. Retryability deliberately still differs and the comment now says why — an unrecognised category is ordinary for an arbitrary OpenAI-shaped gateway, so making only this case permanent would contradict the default arm beside it, whereas anthropic knows its upstream's taxonomy in full | requested by @claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | `EHOSTDOWN`, `ENETDOWN`, `ENETRESET` missing from the errno set: added | requested by @claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Truncated-body test asserting `"unexpected EOF"` could flake to `"connection reset by peer"`: message assertion dropped for that case, Code and Retryable kept. Both shapes pinned exactly in the unit tests instead | requested by @claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Found while checking the above, not raised: an HTTP error whose body is not OpenAI-shaped JSON left `Code` and `Message` empty, so the error's entire text was `server error: `. A proxy's HTML error page is enough. Falls back to the status code and the SDK's own message, with a test | [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | HTTP/2 `GOAWAY` should be retryable: not changed. `net/http` handles the retryable GOAWAY cases itself and reports the residue through the unexported `errClientConnGotGoAway`, which `errors.Is` cannot reach; matching `*http2.GoAwayError` would mean a direct dependency on `x/net/http2` internals for a shape I could not reproduce. The `io.EOF` arm covers the observable "peer hung up" result | answered, @claude |
| 1 | `option.WithMaxRetries(0)` so the interceptor owns retry policy: not changed, and the point is fair — the SDK dials three times before `classifyError` runs, so "reported immediately" was overstated. The PR body now records it. Setting it to 0 changes the retry behaviour every caller of this provider gets today, and the same argument applies to `providers/openai`, `anthropic` and `bedrock`: that is a decision about where retry policy lives, not a bug fix | answered, @claude |
| 1 | Provider-side timeout could be flagged retryable using `ctx.Err() == nil` at the call site: not changed, and you are right that the information is unread rather than lost. It needs `ctx` threaded through `classifyError` and the exported `ResponseMapper.FromProviderError`, which is a signature change answering a different question than this PR. Recorded as a separate change | answered, @claude |
| 1 | `llm.ErrInvalidConfig` for the scheme and certificate cases: dismissed. The classifier cannot generically tell a misconfiguration from any other permanent transport failure, and the only way in is string-matching `"unsupported protocol scheme"` — the exact fragility this PR removes. `transport_error` plus the message names the cause | answered, @claude |

### Verification

`go test ./providers/openaicompat/` and `go test -short ./...` pass; `task license:check` and `task examples:build` pass; `golangci-lint run ./providers/openaicompat/...` is back to the three pre-existing findings CI ignores. `GOOS=windows|darwin go vet` clean.

Each of the seven behaviour changes was checked to be genuinely pinned rather than assumed: reverting any one of them individually breaks tests, and the dropped-connection case breaks both the unit test and the end-to-end one.

Requesting fresh review.

birdayz:
@claude looks good now?

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=3 · P3=0

The patch improves classification but introduces incorrect stream-category handling for specific error codes and adds unguarded syscall constants that break some GOOS builds. These are actionable correctness issues.

Full review comments:

- [P2] Prefer specific stream error codes over broad types — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors.go:260-264
  When a stream frame includes both a broad recognized `type` and a more specific recognized `code` (for example `type: "invalid_request_error"` with `code: "content_policy_violation"` or `"invalid_api_key"`), this branch never checks the raw code because the type already produced a base. The classifier then reports `ErrInvalidInput`/`invalid_request_error`, so callers cannot match the intended policy or auth category even though those codes are in `lookupStreamCategory`.

- [P2] Recognize OpenAI's image policy violation code — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors.go:344-345
  For stream frames whose category is carried only in `code`, OpenAI's documented policy-block value is `image_content_policy_violation` (also mapped in `providers/openai/response_mapper.go`), but this lookup only handles `content_policy_violation`. Such frames fall through to the unknown default and become retryable `ErrServerError` instead of `ErrContentPolicyViolation`, causing permanent safety blocks to be retried.

- [P2] Guard syscall errno table for unsupported GOOS — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors.go:145-150
  On targets where these `syscall` errno constants do not exist, this package-level map makes the package fail to compile; for example `GOOS=plan9` has no `syscall.Errno`, and `GOOS=wasip1` lacks `EHOSTDOWN`. Split the errno allowlist behind OS-specific files or avoid unguarded `syscall` constants so non-Unix/Windows builds keep compiling.

Script done on 2026-08-20 01:21:21+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 2 summary (commits 13ad6e6a..74d3ade7)

Round 2 found three things worth having: a build break I introduced, a misclassification of two error shapes OpenAI actually sends, and a panic in code I added in round 1. All fixed. Everything from round 1 is repeated below so this comment stands alone.

### All items addressed

| Round | Item: Fix | Commit |
|-------|-----------|--------|
| 1 | Bare `io.EOF` under a transport wrapper judged permanent: added to `transportRetryable`, not to the entry condition — a clean end of stream cannot reach the function, which is what makes the arm safe | requested by claude and codex [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Winsock errnos never matched, so every dropped connection on Windows was permanent: table now carries the Winsock analogues alongside the POSIX names, exercised on every platform since the lookup is numeric | requested by claude and codex [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Numeric SSE error codes hand-enumerated with a retryable default, so a permanent 402/404 got hammered: seven string arms replaced by `strconv.Atoi` + `classifyStatusCode` | requested by claude and codex [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Reading only `type` lost OpenAI's own 429 shape (`type: "requests"`, `code: "rate_limit_exceeded"`): both fields consulted | requested by claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Codeless frame reused `unparseable_stream_error` and diverged from anthropic's retryability while the comment claimed parity: separate `unknown_stream_error` constant, and the deliberate retryability difference is now explained in the comment | requested by claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | `EHOSTDOWN`, `ENETDOWN`, `ENETRESET` missing from the errno set: added | requested by claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Truncated-body test could flake between `"unexpected EOF"` and `"connection reset by peer"`: message assertion dropped, both shapes pinned exactly in unit tests instead | requested by claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 1 | Found while checking the above: an HTTP error with a non-OpenAI-shaped body left `Code` and `Message` empty, so the error's whole text was `server error: `. Falls back to the status code and the SDK's message | [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| 2 | **Build break I introduced**: the errno table referenced `syscall` constants unconditionally, which fails to compile where a platform lacks them — plan9 has no `syscall.Errno`, wasip1 is missing `EHOSTDOWN`. plan9 never built this module anyway (franz-go and grpc get there first), but wasip1 did until I broke it. Table and errno arm moved behind `unix \|\| windows` with a stub elsewhere; every GOOS builds and vets | requested by codex [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| 2 | **Wrong precedence between `type` and `code`**: they are not equals — `code` is the specific reason, `type` the family. OpenAI puts `invalid_api_key` and `content_policy_violation` in `code` under type `invalid_request_error`, so reading `type` first reported an auth failure as invalid input and a safety block as an ordinary bad request. Takes the most specific recognised value now, still falling back to the type when the code means nothing to us, so type-only frames are unaffected. This reverses the round-1 precedence and the test that pinned it | requested by codex [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| 2 | `image_content_policy_violation` unrecognised, so a permanent safety block was retried — `llm/errors.go` has documented that code as mapping to `ErrContentPolicyViolation` all along | requested by codex [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| 2 | **Panic in the round-1 fallback**: `apiErr.Error()` formats `Request` and `Response` with no nil check, and `FromProviderError` is exported, so a caller-built `&openai.Error{StatusCode: 500}` panicked where it used to return an empty message. Guarded, with a test that panics against the round-1 code | requested by claude [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| 2 | 408 and 409 classified permanent while openai-go's own `shouldRetry` retries both, so the SDK retried and the interceptor then refused to. Added to `classifyStatusCode`, which the stream path shares | requested by claude [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| 2 | An SSE frame whose `error` member is a bare string, which some OpenAI-compatible servers send, was reported as unparseable — it is readable, it just carries no category. Same for a valid frame with no `error` member at all | requested by claude [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| 2 | `EADDRNOTAVAIL` (ephemeral port exhaustion against one gateway host:port) and `ENOBUFS` missing from the errno set: added, with their Winsock analogues | requested by claude [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| 2 | HTTP status table tested only for the original seven codes: 402, 404, 408, 409, 504 and 529 added, so the function the stream path delegates to is pinned from both directions. Plus a case for a recognised `code` under an unrecognised `type` and vice versa | requested by claude [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| 2 | H2 stream failures and TLS alerts are transient but judged permanent by the allowlist: recorded in `transportRetryable`'s doc comment rather than changed. `net/http` bundles its own http2 and keeps `http2StreamError`/`http2GoAwayError` unexported, so `errors.As` against `x/net/http2` cannot reach them even though the module is already a dependency; `http2.Transport` retries `REFUSED_STREAM` and unprocessed `GOAWAY` itself. Both are distinguishable only by message text, and reading that is the fragility this function exists to remove | answered + documented, claude and codex [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| 2 | Streaming path had no hang-up case, unlike the non-streaming one, which is exactly the asymmetry commit 2 exists to close: added, and the hang-up server is now shared | self [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| 1 | `option.WithMaxRetries(0)` so the interceptor owns retry policy: not changed, and the point is fair — the SDK dials three times before `classifyError` runs, so "reported immediately" was overstated. The PR body records it. Setting it to 0 changes the retry behaviour every caller gets today, and the same argument applies to `providers/openai`, `anthropic` and `bedrock`: that is a decision about where retry policy lives | answered, claude |
| 1 | Provider-side timeout could be flagged retryable via `ctx.Err() == nil` at the call site: not changed, and the observation is right — the information is unread, not lost. It needs `ctx` threaded through `classifyError` and the exported `ResponseMapper.FromProviderError`, a signature change answering a different question | answered, claude |
| 1 | `llm.ErrInvalidConfig` for the scheme and certificate cases: dismissed. The classifier cannot generically tell a misconfiguration from any other permanent transport failure, and the only route in is string-matching `"unsupported protocol scheme"` — the exact fragility this PR removes | answered, claude |
| 2 | `API call failed: API call failed:` double prefix on non-retryable transport errors: not changed. `model.go` double-wraps deliberately, a convention shared with `providers/anthropic`, and it is pre-existing for 401/403. Recorded in the PR body since this PR makes it the normal presentation rather than a rare one | answered, claude |
| 2 | `io.ErrUnexpectedEOF` in the transport entry condition would label a streaming-decoder truncation `transport_error`: not reachable today, since openai-go unmarshals whole buffers and yields `*json.SyntaxError`. Left alone | answered, claude |
| 2 | Unit tests spend seconds in openai-go's internal retry backoff: not changed. `MaxRetries` is a client option with no test-only route in, so cutting it means adding provider surface for a 4s package runtime | answered, claude |

### Verification

`go test` and `go test -short ./...` pass, `-race` clean, 8 consecutive runs of the transport tests with no flake. `task license:check` and `task examples:build` pass. `golangci-lint` is back to the three pre-existing findings CI ignores. Every `GOOS` builds, and `go vet` (which compiles the tests too) is clean on linux, darwin, windows and wasip1.

Each behaviour change was checked to be genuinely pinned rather than assumed: reverting any one of them individually breaks the tests that claim it. The panic guard has a test that panics against the code it replaces. The build fix is verified by cross-compiling every `GOOS`, which no unit test can express.

All previous threads and comments minimized. Requesting fresh review.

birdayz:
@claude looks good now?

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=1 · P3=0

The patch regresses classification for empty successful JSON responses, leaving a real decode failure without the provider error metadata and retry behavior the classifier is meant to provide.

Review comment:

- [P2] Classify wrapped EOFs from empty JSON responses — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors.go:123-123
  When a provider returns 200 with an empty `application/json` body, openai-go's non-streaming path uses `json.Decoder.Decode`, which returns `io.EOF` wrapped as `error parsing response json: EOF` rather than a `json.SyntaxError`. Since this condition now only admits `net.Error` and `io.ErrUnexpectedEOF`, that response falls through unclassified, so callers lose the ProviderError code and retryability that this classifier previously supplied.

Script done on 2026-08-20 01:39:25+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 3 summary (commits 13ad6e6a..5f312b71)

Round 3 found two more instances of the very defect this branch exists to fix, in places I had not looked: the HTTP path retrying a permanent billing failure, and an empty response body reaching the caller with no category at all. Both fixed. Rounds 1 and 2 are repeated so this comment stands alone.

### Round 3

| Item: Fix | Commit |
|-----------|--------|
| **The HTTP path disagreed with the stream path about the same condition.** It classified by status alone, so `insufficient_quota` — which arrives as a 429 — was retried like a rate limit until the caller's budget ran out, while the identical code inside an SSE frame was correctly permanent. And `content_policy_violation` arrives as a 400, so a safety block was indistinguishable from a malformed request. The information was already in hand: `classifyHTTPError` read `apiErr.Code` and put it in the `ProviderError`, it just never asked what it meant. Both paths now share `mostSpecificCategory`, with the status as the fallback; `lookupStreamCategory` loses its `Stream` prefix since it is no longer about streams | requested by claude [`5f312b71`](https://github.com/redpanda-data/ai-sdk-go/commit/5f312b71f8fbc7900a29d08dca1371f09613bb04) |
| **A 200 with an empty `application/json` body was unclassified.** openai-go decodes non-streaming responses with `json.Decoder`, which reports zero bytes as `io.EOF` rather than a syntax error, wrapped as `error parsing response json: EOF` — satisfying neither `net.Error` nor `io.ErrUnexpectedEOF`. It is a short body, not a transport failure, so it gets `response_decode_error` and retries for the same reason a syntax error does. The arm sits after the transport one on purpose: a `*url.Error` wrapping `io.EOF` is a peer that hung up and that arm has to claim it first, and it tests for a *wrapped* `io.EOF` because on its own `io.EOF` is the clean-termination sentinel | requested by codex [`5f312b71`](https://github.com/redpanda-data/ai-sdk-go/commit/5f312b71f8fbc7900a29d08dca1371f09613bb04) |
| `parseStreamErrorFrame` reports `{"error": 503}` or `{"error": []}` as unparseable: not changed. Neither is a shape any upstream is known to send, and inventing a reading for a number would be guessing at whether it is a status or an internal code — the frame's own message still reaches the caller either way. If one turns up, the fix is one more `json.Unmarshal` attempt | answered, claude |
| `lookupCategory` sends any digit-string to `classifyStatusCode` with no range bound, so a four-digit gateway code like `"1301"` hits the `>= 500` arm: not changed. It lands on retryable `ErrServerError`, which is exactly what the unrecognised-category default gives it, so there is no behaviour to fix — only a range check to add for its own sake | answered, claude |

### Round 2

| Item: Fix | Commit |
|-----------|--------|
| **Build break I introduced**: the errno table referenced `syscall` constants unconditionally, which fails to compile where a platform lacks them — plan9 has no `syscall.Errno`, wasip1 is missing `EHOSTDOWN`. plan9 never built this module anyway (franz-go and grpc get there first), but wasip1 did until I broke it. Table and errno arm moved behind `unix \|\| windows` with a stub elsewhere | requested by codex [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| **Wrong precedence between `type` and `code`**: `code` is the specific reason, `type` the family. OpenAI puts `invalid_api_key` and `content_policy_violation` in `code` under type `invalid_request_error`, so reading `type` first reported an auth failure as invalid input. Reverses the round-1 precedence and the test that pinned it | requested by codex [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| `image_content_policy_violation` unrecognised, so a permanent safety block was retried — `llm/errors.go` has documented that code all along | requested by codex [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| **Panic in the round-1 fallback**: `apiErr.Error()` formats `Request` and `Response` with no nil check, and `FromProviderError` is exported, so a caller-built `&openai.Error{StatusCode: 500}` panicked. Guarded, with a test that panics against the round-1 code | requested by claude [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| 408 and 409 classified permanent while openai-go's own `shouldRetry` retries both | requested by claude [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| An SSE frame whose `error` member is a bare string was reported as unparseable — it is readable, it just carries no category. Same for a valid frame with no `error` member | requested by claude [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| `EADDRNOTAVAIL` (ephemeral port exhaustion) and `ENOBUFS` missing from the errno set, with their Winsock analogues | requested by claude [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| HTTP status table tested only for the original seven codes: 402, 404, 408, 409, 504, 529 added | requested by claude [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| H2 stream failures and TLS alerts are transient but judged permanent: recorded in `transportRetryable`'s doc comment rather than changed. `net/http` bundles its own http2 and keeps `http2StreamError`/`http2GoAwayError` unexported, so `errors.As` against `x/net/http2` cannot reach them; `http2.Transport` retries `REFUSED_STREAM` and unprocessed `GOAWAY` itself. Both are distinguishable only by message text | answered + documented, claude and codex [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| Streaming path had no hang-up case, unlike the non-streaming one — exactly the asymmetry commit 2 exists to close | self [`74d3ade7`](https://github.com/redpanda-data/ai-sdk-go/commit/74d3ade74535c725f81144622471127abbf851a7) |
| `API call failed: API call failed:` double prefix: not changed. `model.go` double-wraps deliberately, a convention shared with `providers/anthropic`, and it is pre-existing for 401/403. Recorded in the PR body since this PR makes it the normal presentation | answered, claude |
| `io.ErrUnexpectedEOF` in the entry condition would mislabel a streaming-decoder truncation: not reachable today, since openai-go unmarshals whole buffers and yields `*json.SyntaxError` | answered, claude |
| Unit tests spend seconds in openai-go's internal retry backoff: `MaxRetries` is a client option with no test-only route in, so cutting it means adding provider surface for a 4s package runtime | answered, claude |

### Round 1

| Item: Fix | Commit |
|-----------|--------|
| Bare `io.EOF` under a transport wrapper judged permanent: added to `transportRetryable`, not to the entry condition — a clean end of stream cannot reach the function, which is what makes the arm safe | requested by claude and codex [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| Winsock errnos never matched, so every dropped connection on Windows was permanent | requested by claude and codex [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| Numeric SSE error codes hand-enumerated with a retryable default, so a permanent 402/404 got hammered: replaced by `strconv.Atoi` + `classifyStatusCode` | requested by claude and codex [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| Reading only `type` lost OpenAI's 429 shape (`type: "requests"`, `code: "rate_limit_exceeded"`) | requested by claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| Codeless frame reused `unparseable_stream_error` and diverged from anthropic's retryability while claiming parity: separate `unknown_stream_error`, and the deliberate difference is explained in the comment | requested by claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| `EHOSTDOWN`, `ENETDOWN`, `ENETRESET` missing from the errno set | requested by claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| Truncated-body test could flake between `"unexpected EOF"` and `"connection reset by peer"`: message assertion dropped, both shapes pinned exactly in unit tests | requested by claude [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| Found while checking the above: an HTTP error with a non-OpenAI-shaped body left `Code` and `Message` empty, so the error's whole text was `server error: ` | self [`ce7ff09f`](https://github.com/redpanda-data/ai-sdk-go/commit/ce7ff09fe9316a767f05b88826ff9d3047738c9c) |
| `option.WithMaxRetries(0)` so the interceptor owns retry policy: not changed, and the point is fair — the SDK dials three times before `classifyError` runs, so "reported immediately" was overstated. Setting it to 0 changes what every caller gets today, and the same argument applies to `providers/openai`, `anthropic` and `bedrock` | answered, claude |
| Provider-side timeout could be flagged retryable via `ctx.Err() == nil`: not changed, and the observation is right — the information is unread, not lost. It needs `ctx` threaded through `classifyError` and the exported `FromProviderError` | answered, claude |
| `llm.ErrInvalidConfig` for the scheme and certificate cases: dismissed. The only route in is string-matching `"unsupported protocol scheme"` — the exact fragility this PR removes | answered, claude |

### Verification

`go test` and `go test -short ./...` pass, `-race` clean, no flake over repeated runs of the transport tests. `task license:check` and `task examples:build` pass. `golangci-lint` is back to the three pre-existing findings CI ignores. `go vet` — which compiles the tests too — is clean on linux, darwin, windows, wasip1 and plan9.

Every behaviour change was checked to be genuinely pinned rather than assumed: reverting any one of them individually breaks the tests that claim it. The panic guard has a test that panics against the code it replaces. The build fix is verified by cross-compiling every `GOOS`, which no unit test can express.

All previous threads and comments minimized. Requesting fresh review.

birdayz:
@claude looks good now?

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=1 · P3=0

The new classifier can still retry permanent provider errors when a numeric status in `code` is paired with a more specific semantic `type`. This undermines the intended body-category precedence for quota and policy errors.

Review comment:

- [P2] Do not let numeric status codes mask semantic types — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors.go:382-383
  When an upstream includes both a numeric `code` and a semantic `type`, this treats the numeric HTTP status as a recognized category, so `mostSpecificCategory` returns before considering the type. For example, `{"code":429,"type":"insufficient_quota"}` is classified as retryable rate limiting, reintroducing the quota-retry behavior this patch is trying to avoid; similar numeric 400 responses can also lose `content_policy_violation`.

Script done on 2026-08-20 02:04:02+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 4 summary (commits 13ad6e6a..45fe14ab)

Round 3's fix had a hole that put the defect it fixed straight back, and codex found it. Fixed, along with three of the four optional items from the round-3 review. Rounds 1-3 are in the [round 3 summary](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349420426) and unchanged; only round 4 is listed here.

### Round 4

| Item: Fix | Commit |
|-----------|--------|
| **A status was treated as a category, so it could mask one.** An HTTP status names an outcome; an error code names a reason. Taking them as interchangeable candidates meant `{"code": 429, "type": "insufficient_quota"}` classified as a retryable rate limit — putting back exactly the retry storm on a permanent billing failure that round 3 removed. A status is held back now and only decides once no candidate carries a category. Pinned on both paths | requested by codex [`45fe14ab`](https://github.com/redpanda-data/ai-sdk-go/commit/45fe14abbf2b450fc9d1b13a948e541f43c12d7d) |
| **The taxonomy was missing codes `llm/errors.go` already documents**, and they matter most on the SSE path where there is no status to fall back on — so each one missing was a permanent input error handed the caller's whole retry budget. Azure's `content_filter` is the sharp one: Azure is the commonest non-OpenAI endpoint this provider points at, and without it a safety block there read as an ordinary bad request over HTTP and as a retryable server error mid-stream | requested by claude [`45fe14ab`](https://github.com/redpanda-data/ai-sdk-go/commit/45fe14abbf2b450fc9d1b13a948e541f43c12d7d) |
| **A non-string `message` cost the frame its category.** `json.Unmarshal` rejects the whole object when one field's type disagrees, so `{"error":{"type":"invalid_request_error","message":{...}}}` came out as a retryable unparseable frame despite its own type saying it was permanent. `message` is read raw like `code` now | requested by claude [`45fe14ab`](https://github.com/redpanda-data/ai-sdk-go/commit/45fe14abbf2b450fc9d1b13a948e541f43c12d7d) |
| **The untested quadrant of the body-over-status rule is pinned**: a body saying the request was invalid makes a 5xx permanent. That is the one direction that can cost a retry which used to happen, so it now has a case and a line in the doc comment saying it is intended — the body is the upstream's own statement of what went wrong and a status cannot express it | requested by claude [`45fe14ab`](https://github.com/redpanda-data/ai-sdk-go/commit/45fe14abbf2b450fc9d1b13a948e541f43c12d7d) |
| `net.ErrClosed` added to the transport allowlist — the idle-connection race. Unlike the two documented gaps it is a plain sentinel and needs no message-text reading | requested by claude [`45fe14ab`](https://github.com/redpanda-data/ai-sdk-go/commit/45fe14abbf2b450fc9d1b13a948e541f43c12d7d) |
| PR body overstated that `Code` always names the field that decided the classification: corrected. When nothing is recognised and the status decides, `Code` is the most specific value the frame actually carried, which is what the code comment already said | requested by claude [`45fe14ab`](https://github.com/redpanda-data/ai-sdk-go/commit/45fe14abbf2b450fc9d1b13a948e541f43c12d7d) |
| `mostSpecificCategory` returning an `error` that is a category sentinel rather than a failure reads oddly at the call site: not changed. It matches the shape `classifyStatusCode` has had all along, so changing one means changing both for no behaviour difference | answered, claude |
| `parseStreamErrorFrame` still reports `{"error": 503}` or `{"error": []}` as unparseable: not changed. Neither is a shape any upstream is known to send, and reading a number would be guessing whether it is a status or an internal code | answered, claude |
| `lookupCategory` had no range bound on digit-strings, so a four-digit gateway code hit the `>= 500` arm: now moot. Numeric candidates are only consulted as a status of last resort, and a value nothing recognises lands on the same retryable default either way | answered, claude |

### Verification

`go test` and `go test -short ./...` pass, `-race` clean, four consecutive runs of the transport and empty-body tests with no flake. `task license:check` and `task examples:build` pass. `golangci-lint` is at the three pre-existing findings CI ignores. `go vet` — which compiles the tests — is clean on linux/amd64, linux/arm64, darwin, windows, wasip1, js and plan9.

Every behaviour change in this round was checked to be genuinely pinned: reverting each one individually breaks the test that claims it, including reinstating the numeric-masks-type bug, which breaks a case on each path.

All previous threads and comments minimized. Requesting fresh review.

birdayz:
@claude looks good now?

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=1 · P3=0

The patch compiles, but the new stream taxonomy makes unrecognized stream categories retryable while omitting known permanent API categories. That can regress mid-stream authorization/not-found failures into retry loops.

Review comment:

- [P2] Treat permission stream errors as permanent — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors.go:457-457
  When a stream error frame carries permanent API types like `permission_error` or `not_found_error`, there is no HTTP status code on the stream path to correct the classification. Because this allowlist stops at authentication/quota codes, `classifyStreamError` falls into the `base == nil` unknown branch and marks them as retryable server errors, so callers waste retries on 403/404-style failures that cannot succeed.

Script done on 2026-08-20 02:20:59+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 5 summary (commits 13ad6e6a..0f27193e)

Round 4 introduced a category regression on the shape OpenAI sends most often for a missing API key, and round 5's review caught it. Fixed, along with the two smaller points and a taxonomy gap codex found. Rounds 1-3 are in the [round 3 summary](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349420426) and round 4 in the [round 4 summary](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349556931); only round 5 is listed here.

### Round 5

| Item: Fix | Commit |
|-----------|--------|
| **A family name is not a reason either.** Reading the body's category ahead of the status regressed OpenAI's missing-key 401 — `{"type": "invalid_request_error", "code": null}`. That type covers a missing key, an over-long prompt and a safety block alike, so on its own it says less than the 401 does, and the result was an auth failure reported as invalid input with a caller keying on `ErrAPICall` seeing nothing. Retryability was unaffected, which is why it took a third look to notice. The precedence is three-tiered now and named as such: a specific category, then the HTTP status, then a family name — which only decides where there is no status at all, so a frame carrying nothing but `invalid_request_error` is still permanent | requested by claude [`0f27193e`](https://github.com/redpanda-data/ai-sdk-go/commit/d55c01813c515f0437b98c0980d6f96c7d1f8299) |
| **A number the body echoed is not the response's status.** On the HTTP path a numeric body `code` decided before the real status was ever consulted, so a proxy echoing an inner upstream's number outranked the response's own. The status now comes from the response; a body number stands in for it only on the SSE path, where it is the only one there is | requested by claude [`0f27193e`](https://github.com/redpanda-data/ai-sdk-go/commit/d55c01813c515f0437b98c0980d6f96c7d1f8299) |
| **`permission_error` and `not_found_error` were unrecognised**, though this file already recognised `authentication_error` from the same Anthropic taxonomy and `providers/anthropic` maps all of them. A gateway fronting Anthropic models passes those names straight through, and a frame has no status to correct the classification, so a mid-stream 403 or 404 took the retryable default. `api_error` added alongside them: it landed on the retryable default by accident rather than on purpose | requested by codex [`0f27193e`](https://github.com/redpanda-data/ai-sdk-go/commit/0f27193e207e66c958adfbd446a5077b77e3b2ae) |
| Misplaced doc comment: the `io.EOF`/`io.ErrUnexpectedEOF` paragraph had ended up on `TestTransportRetryable_ClosedConnection` instead of `..._DroppedConnection`, which it describes. Moved | requested by claude [`0f27193e`](https://github.com/redpanda-data/ai-sdk-go/commit/d55c01813c515f0437b98c0980d6f96c7d1f8299) |
| The 5xx pin changed with the precedence: the direction worth pinning is still that a body can make a retryable status permanent, but the case uses a specific category now rather than the family name that no longer qualifies — a 503 whose body names a policy violation, not one that says only "invalid request" | [`0f27193e`](https://github.com/redpanda-data/ai-sdk-go/commit/d55c01813c515f0437b98c0980d6f96c7d1f8299) |

### Verification

`go test` and `go test -short ./...` pass, `-race` clean. `task license:check` and `task examples:build` pass. `golangci-lint` is at the three pre-existing findings CI ignores. `go vet` — which compiles the tests — is clean on linux, darwin, windows, wasip1 and plan9.

Both behaviour changes are mutation-checked: undoing the family-name demotion breaks the 401 case, and letting a body number stand in for the response's status breaks two.

**CI note:** the first `Test` run on `0f27193e` failed in `TestRegistry_WebfetchToolWithLLM_Integration` (`tool/registry_integration_test.go:498`), which asserts a live model's free text contains one of "json", "slideshow" or "data". Nothing to do with this branch — no openaicompat code on that path — and it passed on re-run. Worth a separate look as a flake.

All previous threads and comments minimized. Requesting fresh review.

birdayz:
@claude looks good now?

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=0 · P3=0

No discrete correctness issues were identified in the changed error classification and transport retry handling.

Script done on 2026-08-20 02:39:40+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 6 summary (commits 13ad6e6a..96cfef18)

Codex is at zero findings this round. Claude's review was an approval with two non-blocking points, and both are now fixed — the first because it was the same defect as the rest of this branch, pointing the last remaining way. Earlier rounds: [1-3](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349420426), [4](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349556931), [5](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349707271).

### Round 6

| Item: Fix | Commit |
|-----------|--------|
| **The body could grant a retry the status refused.** Letting the body's category outrank the status was meant to work one way — a specific reason beats an outcome, so a 5xx naming a policy violation stops being retried. It ran both ways: a 401 whose body carried `api_error`, or a 403 carrying `server_error`, came out retryable and got the caller's whole budget. Newly reachable, since `api_error` only became retryable last round. A server that answered 401 is authoritative that this request will not work, whatever its body blames upstream, so the override is now one-way. The clamp applies only where there is a status, so the SSE path is untouched, and 429 with `rate_limit_exceeded` still retries because the 429 itself is retryable. Both directions pinned | requested by claude [`96cfef18`](https://github.com/redpanda-data/ai-sdk-go/commit/96cfef183fee50f17a1ec80737d3953dd44dbcc5) |
| An SSE frame whose `error` member is a bare number landed under `unparseable_stream_error`, the code meant for garbage, when it is a readable frame with no category — the distinction the two codes were split to make. Accepting any scalar covers the bare-string case the same way and is shorter than the string-only check it replaces | requested by claude [`96cfef18`](https://github.com/redpanda-data/ai-sdk-go/commit/96cfef183fee50f17a1ec80737d3953dd44dbcc5) |
| `errnos_syscall_test.go` compiling on Windows, which review could not check: confirmed. `GOOS=windows go vet` compiles the test files, and it is clean, as are darwin, wasip1 and plan9 | verified, claude |

### Where this leaves the branch

Six rounds, and the shape of what they found is worth recording: every round after the first turned up the same defect — retrying something that cannot succeed, or refusing to retry something that can — in a place the previous round had not looked. Transport errnos, then SSE frame categories, then the HTTP path, then the precedence between a reason and a status, then the direction that precedence runs. Two of those were regressions I introduced while fixing the round before, which is the argument for the mutation checking: every behaviour claim in this branch has been verified by reverting it and watching a named test fail.

### Verification

`go test` and `go test -short ./...` pass, `-race` clean. `task license:check` and `task examples:build` pass. `golangci-lint` is at the three pre-existing findings CI ignores (funcorder in `provider.go`, two prealloc in `cache_test.go`) — untouched deliberately, since CI runs `only-new-issues`. `go vet` clean on linux, darwin, windows, wasip1 and plan9. All four CI checks green.

All previous threads and comments minimized.

birdayz:
@claude looks good now?

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=0 · P3=0

The changes compile for the package and the reviewed error-classification logic is consistent with the added tests and existing retry semantics. I did not find a discrete regression that should block the patch.

Script done on 2026-08-20 02:54:21+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 7 summary (commits 13ad6e6a..c2971e2f)

Both reviewers converged: codex reports zero findings for the second round running, and claude's review was an approval with one optional consistency note. That note is answered below with a comment rather than a behaviour change, because the asymmetry it points at is real but load-bearing. Earlier rounds: [1-3](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349420426), [4](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349556931), [5](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349707271), [6](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349771523).

### Round 7

| Item: Fix | Commit |
|-----------|--------|
| The one-way retry clamp is on the HTTP path only, so `{"code": 401, "type": "api_error"}` in a frame stays retryable where the same pair over HTTP is clamped permanent: **deliberate, and now documented** rather than extended. The clamp rests on a status line being the server's own answer, which no payload can contradict. A number the upstream put in its own `"code"` stands in for a status where there is none, but it does not inherit that authority — it is just another field beside the type it lost to, and a frame saying 401 is not the server saying it. Extending the clamp there would give a value that lost the precedence contest a veto over retryability, on the strength of nothing more than being numeric. Recorded in `classifyCategory`'s doc comment so it does not get "fixed" later, the way the `StatusCode == 0` point nearly was | answered + documented, claude [`c2971e2f`](https://github.com/redpanda-data/ai-sdk-go/commit/c2971e2fc09c624b5f40da39ac7c4aedd022b86d) |

Reachability agrees with that reading, for what it is worth: the gateways that send a numeric `code` (OpenRouter) do not send a `type` beside it, so the pair is speculative rather than observed.

### What seven rounds found

Worth recording, because the pattern is the point. Every round after the first turned up the same defect — retrying something that cannot succeed, or refusing to retry something that can — in a place the round before had not looked: transport errnos, then SSE frame categories, then the HTTP path, then the precedence between a reason and a status, then the direction that precedence runs. Three of those were regressions introduced while fixing the round before, which is why every behaviour claim in this branch has been checked by reverting it and watching a named test fail rather than by reading the diff.

### Verification

`go test` and `go test -short ./...` pass, `-race` clean. `task license:check` and `task examples:build` pass. `golangci-lint` is at the three pre-existing findings CI ignores (funcorder in `provider.go`, two prealloc in `cache_test.go`), deliberately untouched since CI runs `only-new-issues`. `go vet` — which compiles the test files too — is clean on linux, darwin, windows, wasip1 and plan9. All four CI checks green.

All previous threads and comments minimized.

birdayz:
@claude looks good now?

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=1 · P3=0

The patch mostly improves classification, but a malformed-yet-readable stream error can now be misclassified as retryable and lose its provider code. This should be fixed before considering the patch fully correct.

Review comment:

- [P2] Preserve stream codes when type is not a string — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors.go:230-230
  When a streaming error object has a non-string `type` but a meaningful `code` (for example `{"error":{"type":401,"code":"invalid_api_key"}}`), unmarshalling into this `string` field returns an error for the whole payload. `parseStreamErrorFrame` then takes the unparseable-frame path and `classifyStreamError` reports a retryable server error, losing the permanent code.

Script done on 2026-08-20 03:06:53+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 8 summary (commits 13ad6e6a..5f4d3cda)

Codex: zero findings, third round running. Claude: approval with one optional item, fixed below. Earlier rounds: [1-3](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349420426), [4](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349556931), [5](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349707271), [6](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349771523), [7](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349839181).

### Round 8

| Item: Fix | Commit |
|-----------|--------|
| **Not every digits-only code is an HTTP status.** `classifyCategory` promoted anything `strconv.Atoi` accepted, and gateways carry their own numeric taxonomies in that field — Zhipu's `1002` for a bad token, `1113` for exhausted quota. Read as a status those sailed past `classifyStatusCode`'s `>= 500` arm and came out retryable: the right answer for the wrong reason, and one that would have followed that arm anywhere it moved. A negative or two-digit vendor code went the other way and became a permanent `ErrAPICall` off a nonsense number. Out of range they fall through to the unrecognised-category arm now, which is what a code we cannot read actually is | requested by claude [`5f4d3cda`](https://github.com/redpanda-data/ai-sdk-go/commit/5f4d3cda60ee7922976cf6406045e72615d39dc9) |

Worth being precise about what that fixes, since it is easy to overstate: every status in the tests (402, 404, 408, 429, 503, 529) is inside the range, and `1002` came out retryable before and comes out retryable now. What changes is the *reason* — an unrecognised vendor code is retryable because openaicompat's policy for categories it does not know is retryable, not because 1002 was mistaken for a server error — plus the nonsense-number cases, which did move. So the guard's contract is pinned directly on `classifyCategory` rather than through a frame that would pass either way; reverting the one condition breaks six cases in `TestClassifyCategory_OnlyStatusRangeStandsInForAStatus`.

### Verification

`go test` and `go test -short ./...` pass, `-race` clean. `task license:check` and `task examples:build` pass. `golangci-lint` is at the three pre-existing findings CI ignores. `go vet` clean on linux, darwin, windows, wasip1 and plan9. All four CI checks green.

**Flake, not this branch:** `Test` failed twice across this branch's runs in `TestRegistry_WebfetchToolWithLLM_Integration` (`tool/registry_integration_test.go:498`), which asserts a live model's free text contains one of "json", "slideshow" or "data". No openaicompat code on that path, and it passes on re-run both times. Worth a separate fix — an assertion on a model's prose is not a test.

All previous threads and comments minimized.

birdayz:
@claude looks good now?

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=1 · P3=0

The stream error parser can discard an otherwise usable error code when only the type field has an unexpected JSON shape, causing incorrect retry classification for those frames.

Review comment:

- [P2] Preserve code when stream error type is malformed — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors.go:230-230
  If an SSE error frame has a non-string `type` value next to a valid `code` (for example `{"error":{"type":123,"code":"invalid_api_key"}}`), this `string` field makes `json.Unmarshal` fail for the entire payload. `parseStreamErrorFrame` then treats the frame as unparseable, so valid permanent codes are reported as retryable server errors instead of being classified from `code`; keep `type` raw/parse independently as with `Code` and `Message`.

Script done on 2026-08-20 03:17:25+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 9 summary (commits 13ad6e6a..fb1efb28)

Claude approved with no correctness findings. Both remaining items — its optional nit and codex's one P2 — are fixed, because both were the same kind of loose end rather than new questions. Earlier rounds: [1-3](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349420426), [4](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349556931), [5](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349707271), [6](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349771523), [7](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349839181), [8](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349937190).

### Round 9

| Item: Fix | Commit |
|-----------|--------|
| **The frame's `type` was the last plain string in `streamErrorPayload`**, and it had exactly the problem `message` had two rounds ago: `json.Unmarshal` rejects the whole object when one field's type disagrees, so `{"type": 123, "code": "invalid_api_key"}` came out as a retryable unparseable frame with a perfectly readable permanent code sitting inside it. All three fields are raw now, which is the state the struct should have been in when the first one moved | requested by codex [`fb1efb28`](https://github.com/redpanda-data/ai-sdk-go/commit/fb1efb285bcee84c1941aa8f03a51e845c09f04f) |
| **A family name that decided the category was also what got reported**, discarding a more specific unrecognised code beside it — while the status branch already did the opposite. `ErrInvalidInput` already says "invalid input", so repeating `invalid_request_error` in the one diagnostic field spends it on the less useful of the two values the frame carried. Both branches report the most specific name present now | requested by claude [`65e8e169`](https://github.com/redpanda-data/ai-sdk-go/commit/65e8e169cac060666269f72ff31fb7177b2f5365) |

### Verification

`go test` and `go test -short ./...` pass, `-race` clean. `task license:check` and `task examples:build` pass. `golangci-lint` is at the three pre-existing findings CI ignores. `go vet` clean on linux, darwin, windows, wasip1 and plan9. All four CI checks green, first attempt.

Both changes mutation-checked: putting `type` back to a plain string breaks the non-string-type case, and reporting the family again breaks the specific-name case.

All previous threads and comments minimized.

birdayz:
@claude looks good now?

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=1 · P3=0

The implementation compiles and the core classifier tests pass locally, but the newly added transport test contains an OS-specific message assertion that will fail on Windows. Because this package explicitly includes Windows-specific retry handling, the cross-platform test failure should be fixed.

Review comment:

- [P2] Avoid asserting Linux-only refused-connection text — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errors_test.go:811-811
  On Windows, the refused-connection error from `net/http` is reported as a Winsock message such as `connectex: No connection could be made because the target machine actively refused it`, not `connection refused`. Since this test is not build-tagged away from Windows, `TestTransportFailures/refused connection may come back` will fail there even though the retry classification is correct; assert only the retry/code behavior or make the message check platform-specific.

Script done on 2026-08-20 03:29:50+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 10 summary (commits 13ad6e6a..38cf5eae)

Claude: approve, no findings. Codex: one P2, fixed below. Earlier rounds: [1-3](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349420426), [4](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349556931), [5](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349707271), [6](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349771523), [7](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349839181), [8](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349937190), [9](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5350022878).

### Round 10

| Item: Fix | Commit |
|-----------|--------|
| **Two transport cases asserted the platform's wording rather than Go's.** `"connection refused"` is POSIX; Windows says `"connectex: No connection could be made because the target machine actively refused it"`, and the x509 wording differs there too since Windows verifies through the system store. CI is Linux-only so neither ever failed, but the tests carry no build tag and would fail for anyone running them elsewhere. Both assert on the wrapper Go itself writes now — `"dial tcp"` and `"tls: failed to verify certificate"` — which is stable everywhere and still identifies the failure being driven. Same reasoning as dropping the truncated-body message assertion in round 1: the classification is the claim, and a message is only worth asserting where we own the words | requested by codex [`38cf5eae`](https://github.com/redpanda-data/ai-sdk-go/commit/38cf5eaea147028f76c07b545e97729cf157035b) |
| `classifyCategory`'s doc said "a number the upstream put in its own `code`" stands in for a status, but since the frame's `type` became raw a number in either candidate does. Corrected | requested by claude [`38cf5eae`](https://github.com/redpanda-data/ai-sdk-go/commit/38cf5eaea147028f76c07b545e97729cf157035b) |

### Verification

`go test` and `go test -short ./...` pass, `-race` clean. `task license:check` and `task examples:build` pass. `golangci-lint` at the three pre-existing findings CI ignores. `go vet` clean on linux, darwin, windows, wasip1 and plan9. All four CI checks green, first attempt.

The reworded assertions still do their job: forcing `transportRetryable` to `return true` breaks five cases across the two transport tests, same as before.

All previous threads and comments minimized.

birdayz:
@claude looks good now?

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=1 · P3=0

The new socket errno allowlist misses a Unix counterpart for a Windows condition it explicitly handles, causing some transient connection shutdowns to be marked permanent. The rest of the reviewed changes appear coherent.

Review comment:

- [P2] Include Unix ESHUTDOWN in retryable socket errors — /home/johannes/projects/ai-sdk-go/worktrees/transport-retry/providers/openaicompat/errnos_syscall.go:34-34
  When a Unix socket write fails with `ESHUTDOWN` (`cannot send after transport endpoint shutdown`), this allowlist falls through even though the Windows analogue is included as `WSAESHUTDOWN` at `10058`. That regresses those dropped-connection failures from the previous retryable transport classification to a permanent `ErrAPICall`, so add `syscall.ESHUTDOWN` alongside `EPIPE`.

Script done on 2026-08-20 03:44:22+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 11 summary (commits 13ad6e6a..cea04bea)

Claude: approve, no blocking findings. Codex: one P2, fixed. Earlier rounds: [1-3](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349420426), [4](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349556931), [5](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349707271), [6](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349771523), [7](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349839181), [8](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5349937190), [9](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5350022878), [10](https://github.com/redpanda-data/ai-sdk-go/pull/202#issuecomment-5350116235).

### Round 11

| Item: Fix | Commit |
|-----------|--------|
| `ESHUTDOWN` missing while `WSAESHUTDOWN` was present, so a write to a shut-down endpoint was retryable on Windows and permanent on Unix. An inconsistency in the table rather than a judgement about the condition, which is why it is worth a line | requested by codex [`cea04bea`](https://github.com/redpanda-data/ai-sdk-go/commit/cea04beab68f360403aa910d3044eadd4955777f) |
| `errors.Unwrap(err) != nil` on the short-body arm tests "has a single-error `Unwrap`" rather than "is wrapped", so an `errors.Join` or multi-`%w` chain carrying `io.EOF` would fall through: left alone. openai-go wraps the decoder's `io.EOF` with a single `%w`, so nothing in the provider can produce that shape; it would take a caller hand-building one and passing it to the exported `FromProviderError`. Widening the guard means either an `err != io.EOF` identity comparison, which `errorlint` rejects, or walking the tree by hand for a case that does not occur | answered, claude |
| `"dial tcp"` would also match a dial timeout, making that row less specific about which transport failure it drives: accepted trade. The row's claim is `wantRetry` plus `llm.IsRetryable`, and the alternative was asserting Winsock-specific prose | answered, claude |

### Where this ends up

Eleven rounds. The tally is worth recording because it is the argument for having run the loop this far rather than stopping at the first approval: **17 defects found after the original two commits**, of which 5 were regressions introduced while fixing an earlier round — a build break on wasip1, a panic reachable through exported API, an auth failure reclassified as invalid input, a retry storm reinstated on a permanent billing failure, and a body category granting a retry a 4xx had refused. Every one is the same underlying mistake the branch exists to fix: retrying what cannot succeed, or refusing to retry what can.

That is also why no behaviour claim here rests on reading the diff. Each one was checked by reverting it and watching a named test fail, and the two claims no unit test can express — the cross-`GOOS` build and the panic — were checked by cross-compiling every `GOOS` and by a test that panics against the code it replaces.

### Verification

`go test` and `go test -short ./...` pass, `-race` clean. `task license:check` and `task examples:build` pass. `golangci-lint` at the three pre-existing findings CI ignores (funcorder in `provider.go`, two prealloc in `cache_test.go`), deliberately untouched since CI runs `only-new-issues`. `go vet` — which compiles the test files — clean on linux/amd64, linux/arm64, darwin, windows, freebsd, wasip1, js and plan9. All four CI checks green.

All previous threads and comments minimized.

birdayz:
@claude looks good now?

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **bedrock** (model `openai.gpt-5.5`). These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=0 · P3=0

The changes compile and the non-network classification tests pass. I did not identify any discrete regressions in the new error classification or platform-specific errno handling.

Script done on 2026-08-20 04:00:37+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
## Round 12: both reviewers clean

Claude: approve, no blocking findings. Codex: P1=0 P2=0 P3=0, second clean round.

One optional nit from claude, answered rather than changed: `classifyCategory`'s three return sites report `Code` differently — the recognised-category arm returns the candidate that decided, while the status and family arms return the most specific name present. That divergence is the rule the doc comment states, not an accident: a recognised category earns the field because it is the reason the request failed, while a status and a family name both leave their meaning in the sentinel and so cede the field to whatever is more specific. `{"code":"tokens","type":"rate_limit_error"}` reporting `rate_limit_error` is the intended answer — `tokens` names the limit dimension, which is opaque on its own, and it is in the message. Pinned by a test that says so.

providers/openaicompat/errors.go
```diff
@@ -102,37 +155,455 @@ func classifyNonHTTPError(err error) error {
...
+	// load balancer or a keep-alive race) or part-way through the body. The
+	// request itself may have been fine.
+	//
+	// Bare io.EOF is also the conventional clean-termination sentinel, which is
+	// why classifyNonHTTPError keeps it out of the entry condition. A clean end
+	// of stream therefore cannot reach this function: it is not a net.Error and
+	// not io.ErrUnexpectedEOF, so it is returned unclassified before we get
+	// here. Inside a net.Error wrapper, io.EOF only ever means a dropped
+	// connection.
+	return errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, io.EOF)
```
malinskibeniamin: **Priority: P1**

**Transport allowlist turns transient HTTP/2 stream failures into permanent errors.**

`transportRetryable` is an allowlist ending in this expression, so anything it does not name returns `false`. Before this PR, `classifyNonHTTPError` marked every `net.Error` retryable, so an HTTP/2 `RST_STREAM` / mid-response `INTERNAL_ERROR` (and a load-shedding `remote error: tls: handshake failure`) arrived wrapped in `*url.Error`, matched `net.Error`, and was retried. Now neither matches any arm here — not `Timeout()`, not `*net.DNSError`, not `transientErrno` (the errno is never a `syscall.Errno`), not `net.ErrClosed`, not `io.EOF`/`io.ErrUnexpectedEOF` — so it comes out `llm.ErrAPICall` with `Retryable: false`.

The doc comment at lines 171–180 acknowledges this deliberately, but the consequence is user-facing: h2 is the default for `https://` upstreams, and OpenAI/OpenRouter-class gateways emit server-initiated stream resets routinely. A caller's configured retry budget is now bypassed and an intermittent request fails hard, which is exactly the class of failure retries exist for. The stated blocker (net/http vendors http2 and keeps `http2StreamError`/`http2GoAwayError` unexported, so `errors.As` against `golang.org/x/net/http2` cannot match) is real, but it argues for a different default, not for classifying these as permanent.

Correction: keep the allowlist for the *permanent* decisions it was built to make and default the remainder to retryable — i.e. reject the known-permanent shapes explicitly (unsupported scheme, `*tls.CertificateVerificationError` / `x509.UnknownAuthorityError`, `*net.DNSError` with `IsNotFound`) and return `true` for everything else that reached the transport arm. That preserves the misconfiguration fix this function was added for (a scheme typo and an untrusted cert still fail fast, as `TestTransportFailures` pins) without silently downgrading every transport failure the taxonomy does not enumerate. If the allowlist default must stay, add narrow recognition for h2 resets — the vendored types are unexported, but their messages are stable (`http2: stream closed`, `INTERNAL_ERROR`, `GOAWAY`) and a scoped string match here is a smaller cost than losing retries on the default protocol.

Verify: add a case to `TestTransportFailures` driving `httptest.NewUnstartedServer` with `EnableHTTP2 = true` whose handler panics mid-response (yielding a stream reset), assert `pe.Retryable` and `llm.IsRetryable(err)` are true, then `go test ./providers/openaicompat/ -run TestTransportFailures -count=1`.

*Automated `/review`.*

malinskibeniamin (commented):

providers/openaicompat/errors.go
birdayz: Right diagnosis, and the correction you proposed is what landed — but the mechanism is worse than described, and it argues for going further than the inversion.

An h2 reset never reached `transportRetryable` at all. It arrives as net/http's `http2StreamError`, which is not a `net.Error`, so it failed the *entry condition* to the transport arm and fell out of the classifier with no code, no category and `Retryable` unset. Not "came out ErrAPICall with Retryable: false" — no `*llm.ProviderError` in the chain whatsoever. That was already true on main before any allowlist existed, so it is a pre-existing gap this PR was making harder to see rather than a regression it introduced.

So inverting to a denylist would not have fixed it. I tried that first: default-retryable plus explicit rejects for unsupported scheme, `*tls.CertificateVerificationError`/`x509` and `IsNotFound` DNS. The h2 test still failed, because widening the *default* does nothing for an error that never enters the arm. Fixing it needed the entry condition widened too, at which point I was string-matching `stream error: stream ID`, `http2: stream closed` and `GOAWAY` — your fallback suggestion — to reach a case that a denylist over "everything else" catches for free.

Which is where the whole thing collapsed. If unrecognised transport failures default to retryable, and the entry condition is "everything else", then the denylist buys exactly one thing: skipping ~1.4s of retries (`plugins/retry`: 3 attempts at 200/400/800ms) on a misconfiguration that then reports the same error anyway. That is not worth the machinery it costs — per-platform errno tables and their build-tagged files, x509 matching, two string matches. It also had a live bug of its own: an errno table classifies every socket failure permanent on Windows, since `WSAECONNREFUSED` is not `syscall.ECONNREFUSED`.

So: no test for what kind of transport failure it is. Anything unrecognised is a retryable `transport_error`. The asymmetry is the argument — 1.4 wasted seconds versus a dead task. The branch is now one commit, +110/-29, down from 1724 insertions across five files, and the errno files going away also removes the `wasip1`/`plan9` build breakage they had introduced.

Kept from your comment: the `UnmarshalTypeError` split (well-formed JSON disagreeing with our structs replays identically, so it does not retry) and your verification recipe. `TestHTTP2StreamResetIsRetryable` drives a real reset and fails against main. One note for anyone reusing it — it has to abort with `panic(http.ErrAbortHandler)`; a short `Content-Length` yields a plain EOF and passes even with the fix disabled, which I only caught by checking.

weeco (approved):

PR #203, opened 2026-08-20
Make context-window overflow identifiable and testable

weeco:
## Why

We're building context compaction for the agent loop. Before it can exist, two gaps need closing:

- A context-window overflow is unrecognisable in code. It arrives as a generic `ErrInvalidInput`, so nothing can react to it (e.g. compact and retry).
- It's untestable. Real windows are 200K+ tokens, so reproducing an overflow is too slow and expensive to iterate on.

## What

- `llm.ErrContextWindowExceeded`, mapped from each provider's prompt-too-long rejection. Wraps `ErrInvalidInput`, so existing error handling keeps working.
- `fakellm` can now enforce a small declared window, making overflow reproducible in milliseconds.
- A property test in `llmagent` pinning the guarantee compaction will be held to: the model never receives an over-window request.
- Drive-by: deflaked the webfetch LLM integration test.

No behaviour change for existing users. Compaction itself comes next.

### DIFF
--- a/llm/errors.go
+++ b/llm/errors.go
@@ -62,23 +62,17 @@ var (
 	// Note: Soft filtering (incomplete response) uses FinishReasonContentFilter instead.
 	ErrContentPolicyViolation = errors.New("content policy violation")
 
+	// ErrContextOverflow indicates the request did not fit the model's
+	// context window. Not retryable until the conversation is shortened.
+	// Wraps ErrInvalidInput. Distinct from FinishReasonContextOverflow, which
+	// is a completed call whose response may already carry content.
+	ErrContextOverflow = fmt.Errorf("%w: context window exceeded", ErrInvalidInput)
+
 	// ErrServerError indicates a transient provider-side error.
 	// Retryable with backoff.
 	// Maps from errors such as: "server_error", "vector_store_timeout", "image_parse_error",
 	//   "failed_to_download_image", "image_file_not_found"
 	ErrServerError = errors.New("server error")
-
-	// ErrContextWindowExceeded indicates the request was rejected before generation
-	// because the input — or the input plus the requested max_tokens — does not fit
-	// the model's context window. Not retryable as-is: the conversation must be
-	// shortened or restarted. A more specific case of ErrInvalidInput, distinct so
-	// callers can give the user an actionable "conversation too long" message. This
-	// is the pre-generation rejection; FinishReasonContextOverflow is the same
-	// condition surfaced on a 200 response.
-	//
-	// It wraps ErrInvalidInput so callers that branch on the coarser category keep
-	// matching these 400s.
-	ErrContextWindowExceeded = fmt.Errorf("context window exceeded: %w", ErrInvalidInput)
 )
 
 // ProviderError wraps a sentinel error with provider-specific details.
@@ -96,8 +90,9 @@ var (
 //	    log.Printf("Provider error [%s]: %s", perr.Code, perr.Message)
 //	}
 type ProviderError struct {
-	// Base is the sentinel error indicating the category
-	// (ErrRateLimitExceeded, ErrInvalidInput, etc.)
+	// Base is the sentinel error indicating the category (ErrInvalidInput,
+	// ErrRateLimitExceeded, ...) or a refinement of one such as
+	// ErrContextOverflow. Match it with errors.Is, not ==.
 	Base error
 
 	// Code is the provider-specific error code ("rate_limit_exceeded", etc.)
--- a/llm/fakellm/fake.go
+++ b/llm/fakellm/fake.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"hash/fnv"
 	"iter"
+	"slices"
 	"sync"
 	"testing"
 	"time"
@@ -43,6 +44,13 @@ type FakeModel struct {
 	tokenizer Tokenizer
 	latency   LatencyProfile
 	defaults  defaults
+
+	// constraints is what Constraints() reports; WithContextWindow shrinks
+	// MaxInputTokens to make overflow reachable.
+	constraints llm.ModelConstraints
+
+	// enforceWindow rejects requests larger than constraints.MaxInputTokens.
+	enforceWindow bool
 
 	// Conversation tracking
 	sessionKeyFrom func(*llm.Request) string
@@ -123,6 +131,10 @@ type Call struct {
 	// Err is any error that was returned
 	Err error
 
+	// InputTokens is the request's counted size, recorded when the call was
+	// made rather than derived from Request afterwards.
+	InputTokens int
+
 	// RuleName is the name of the rule that matched (empty if no rule matched)
 	RuleName string
 }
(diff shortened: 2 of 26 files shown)


llm/fakellm/fake.go
```diff
@@ -287,6 +318,12 @@ func (m *FakeModel) GenerateEvents(ctx context.Context, req *llm.Request) iter.S
 
 	cc := m.beginCall(req)
 
+	// See Generate. A rejection surfaces as a terminal StreamEndEvent.Error, not
+	// an iterator error, so the stream shape stays consistent.
+	if err := m.checkContextWindow(req, CallGenerateEvents); err != nil {
+		return rejectStream(m, cc, err)
```
malinskibeniamin, line 325: **Priority: P1**

The fake reports an over-window rejection through `StreamEndEvent{Error: ...}`, but every real provider surfaces a request-time API failure as an **iterator** error wrapped in `llm.ErrAPICall` — `providers/anthropic/model.go:246`, `providers/openai/model.go:180`, `providers/google/model.go:126`, `providers/bedrock/model.go:115`, `providers/openaicompat/model.go:217` — and so does the fake's own `ThenError` (`llm/fakellm/fake.go:1029`, `yield(nil, err)`).

Consequence, concretely: `plugins/retry`'s streaming path inspects only iterator errors (`plugins/retry/retry.go:180-193`); a `StreamEndEvent.Error` is forwarded as a plain event and never reaches `shouldRetry`. A fake-based test of retry/backoff or any interceptor that classifies stream failures will pass while the production path takes a different branch. The fake's error also omits the `llm.ErrAPICall` wrapper that all five providers add, so a caller matching `ErrAPICall` behaves differently against the fake than in production. `agent/llmagent/llmagent.go:455` and `:486` happen to converge, which is why the new `agent/llmagent/context_budget_test.go` assertions still hold — that convergence hides the divergence rather than justifying it.

Correction: emit the rejection as an iterator error, matching providers and `ThenError`:

```go
func rejectStream(m *FakeModel, cc *CallContext, err error) func(func(llm.Event, error) bool) {
	return func(yield func(llm.Event, error) bool) {
		defer m.endCall(cc)

		yield(nil, fmt.Errorf("%w: %w", llm.ErrAPICall, err))
	}
}
```

and update `llm/fakellm/context_window_test.go:151-172` to expect the error on the iterator. If the `StreamEndEvent` shape is intentional, please state which provider produces it, because none in this repo does.

Verify: `go test ./llm/fakellm/... ./plugins/retry/... ./agent/llmagent/...`

*Automated `/review`.*

malinskibeniamin (commented):

malinskibeniamin (commented):
Automated `/review`: 2 finding(s).

llm/errors.go
```diff
@@ -62,23 +62,17 @@ var (
 	// Note: Soft filtering (incomplete response) uses FinishReasonContentFilter instead.
 	ErrContentPolicyViolation = errors.New("content policy violation")
 
+	// ErrContextOverflow indicates the request did not fit the model's
+	// context window. Not retryable until the conversation is shortened.
+	// Wraps ErrInvalidInput. Distinct from FinishReasonContextOverflow, which
+	// is a completed call whose response may already carry content.
+	ErrContextOverflow = fmt.Errorf("%w: context window exceeded", ErrInvalidInput)
```
malinskibeniamin, line 69: **Priority: P1**

**Exported sentinel removed without a compatibility alias.**

`origin/main` already exports `ErrContextWindowExceeded` (added in #185, commit `2e4d333`). This diff deletes it and introduces `ErrContextOverflow` in its place. Every in-repo reference was updated (`adapter/a2a/executor.go`, `llm/api.go`, the provider classifiers — verified by search, no stragglers), so the only breakage is downstream: any consumer of this module that matches the old sentinel now fails to build with `undefined: llm.ErrContextWindowExceeded`. A rename of an exported identifier in a published Go module is a hard, silent-until-upgrade break, not a soft one.

**Correction:** keep a deprecated alias next to the new sentinel so existing `errors.Is` call-sites keep matching:

```go
// Deprecated: use ErrContextOverflow.
ErrContextWindowExceeded = ErrContextOverflow
```

(Identity assignment, not a second `fmt.Errorf`, so `errors.Is` succeeds in both directions.)

**Verify:** `go build ./... && go test ./llm/... ./adapter/a2a/...`, plus a scratch module that references `llm.ErrContextWindowExceeded` compiled against this commit.

*Automated `/review`.*

providers/openai/errors.go
```diff
@@ -49,6 +77,81 @@ func classifyHTTPError(apiErr *oai.Error) *llm.ProviderError {
...
+	errType, code, message := payload.Type, payload.Code, payload.Message
+	if payload.Error.Code != "" || payload.Error.Message != "" {
+		errType, code, message = payload.Error.Type, payload.Error.Code, payload.Error.Message
+	}
+
+	if code == "" && message == "" {
+		return nil
+	}
+
+	base, retryable := llm.ErrServerError, true
```
malinskibeniamin: **Priority: P1**

**Unrecognized SSE errors now default to retryable, reversing prior behavior.**

Before this diff, `classifyError` did not handle `*ssestream.StreamError` at all: it returned the raw error, and `llm.IsRetryable` (`llm/errors.go:115`) reported `false` for it because it is neither a `*ProviderError` nor a wrapper of `ErrServerError`/`ErrRateLimitExceeded`. The new `classifyStreamError` returns a `*llm.ProviderError` for any payload carrying a `code` or `message`, and this line makes the fall-through case `ErrServerError` with `Retryable: true`.

Only three shapes are recognized: overflow, `errType == "invalid_request_error"`, and rate limits. The Responses API `error` SSE event carries codes such as `insufficient_quota`, `invalid_prompt`, and content-policy/auth failures — none of which match those branches, so each is now reported to callers as a transient server error and handed to the retry plugin. The consequence is a retry storm against a request that can never succeed, plus a wrong error category surfaced to the caller (`ErrServerError` instead of an input/auth/policy failure). The existing tests do not cover this: all four cases in `TestClassifyStreamError` hit a recognized branch.

`providers/openaicompat/errors.go:185` is the same code with the same defect, and there it also shadows the previous `classifyNonHTTPError` fall-through.

**Correction:** default to non-retryable and let only known-transient types opt in:

```go
base, retryable := llm.ErrAPICall, false

switch {
case isContextOverflow(code, message):
	base = llm.ErrContextOverflow
case errType == "invalid_request_error":
	base = llm.ErrInvalidInput
case code == "rate_limit_exceeded" || errType == "rate_limit_error":
	base, retryable = llm.ErrRateLimitExceeded, true
case errType == "server_error" || code == "server_error":
	base, retryable = llm.ErrServerError, true
}
```

Apply the same change in `openaicompat`, and add a table case for an unrecognized code (for example `insufficient_quota`) asserting `IsRetryable == false`.

**Verify:** `go test ./providers/openai/... ./providers/openaicompat/... ./plugins/retry/...`

*Automated `/review`.*

c-julin (approved):

splittingfield (approved):

malinskibeniamin (commented):
Automated `/review`: No P0/P1 findings.

PR #204, opened 2026-08-22
fix(anthropic): put the cache breakpoint on the last cacheable block

birdayz:
## What

The Anthropic request mapper placed the message-side `cache_control` breakpoint on the last *text* block of the last message. It now places it on the last block that can carry a breakpoint, whatever type that block is.

## Why

In an agentic loop the last message is usually a user turn holding nothing but `tool_result` blocks. It has no text block, so the loop marked nothing and the request went out with a breakpoint on the system blocks alone. The static tools+system prefix cached; the conversation never did, and every turn after a tool call re-paid full input price for the entire history.

Measured with a proxy between the agent and the Anthropic API, recording per-call usage and the `cache_control` markers actually present in each request body. The signature is a constant `cache_read` while `input` climbs:

```
single agent:  #2 input  535  cache_read 7881   markers: system
               #3 input  910  cache_read 7881   markers: system
               #4 input 1427  cache_read 7881   markers: system
subagent:      #3 input  541  cache_read 4908   markers: system
               #4 input  916  cache_read 4908   markers: system
               #5 input 1473  cache_read 4908   markers: system
```

First turns, where the last message is text, show `markers: system+messages` and do get a cache write. Every turn after a tool call shows `system` only.

After this change, the same shape measured against the live API by the new integration test in CI:

```
turn 1: input 3, cache_read    0, cache_write 2878
turn 2: input 5, cache_read 2878, cache_write  637
turn 3: input 5, cache_read 3515, cache_write  638
turn 4: input 5, cache_read 4153, cache_write  638
```

`cache_read` now grows with the conversation (2878 -> 3515 -> 4153) and each turn writes only its own ~638-token increment. Before, `cache_read` would have stayed pinned at 2878 — the system prefix — with the growing history re-billed as fresh input every turn.

## Implementation details

**Which blocks get marked.** `markLastCacheableBlock` walks the last message's blocks backwards and marks the first one that can carry `cache_control`. The pinned `anthropics/anthropic-sdk-go` v1.27.1 exposes a `CacheControl` field on every `Beta*BlockParam` in the content union except `BetaThinkingBlockParam` and `BetaRedactedThinkingBlockParam`, which matches the API: text, `tool_use`, `tool_result`, image and document accept a breakpoint, thinking blocks do not.

Only text, `tool_use`, `tool_result` and thinking are reachable from this mapper today. Image and document are covered anyway, because wiring either into `mapUserMessage` later would otherwise drop the breakpoint with no compile error and no failing test — the only symptom being a `cache_read` that quietly stops growing, i.e. this exact bug again. Anything still unrecognized is skipped rather than marked: a 400 on the whole request is a worse outcome than one turn going uncached. A thinking-only last message therefore goes uncached, which is deliberate and now pinned by a test.

The `tool_use` arm is defensive rather than exercised: Anthropic rejects an unanswered `tool_use`, so an assistant turn ending in one cannot legally be the last message. Its correctness rests on the SDK field inventory above, not on a live-API observation.

**Why the marker stays on the tail, by index.** CrewAI fixed the same class of bug by identifying the message to mark by content rather than by index. That reasoning does not transfer here, and the index is deliberate. Anthropic reads the longest already-cached prefix and writes a fresh entry at each breakpoint, so a marker on the newest turn extends the cached prefix by one turn per request — the standard incremental pattern for a growing conversation. Anchoring by content to a stable earlier message would instead re-cache the preamble, which the system-block marker immediately above already covers. The fragility CrewAI was avoiding belongs to the shared-prefix case (many requests over one fixed preamble, breakpoint at the end of the shared part), which is not this code path.

**Bedrock is unaffected.** `providers/bedrock/request_mapper.go` appends a standalone `CachePointBlock` to the last message's content instead of decorating an existing block, so it has no block-type filter to get wrong. No change was needed there.

**What the new tests establish.** `cache_tool_result_test.go` pins the request shape: a `tool_result`-only turn is marked, a parallel batch is marked exactly once on its final block, a trailing thinking block is skipped in favour of the block before it, a thinking-only turn is left unmarked while the system breakpoint survives, and the marker survives JSON serialization onto the `tool_result` block itself. Three of them fail against the old mapper. `cache_conversation_test.go` is the part that separates "the marker is set" from "the conversation actually caches": it replays a growing tool-calling conversation against the live API and asserts `cache_read` grows turn over turn. It also catches a rejected breakpoint, since that arrives as a 400 on `Generate`. The helper in `cache_default_test.go` had the same text-only blind spot as the mapper and would have reported "no marker" for a correctly marked `tool_result` turn, so it was widened too.

**Known residual limit, not addressed here.** Anthropic walks back at most 20 content blocks from a breakpoint to find a prior cache entry, so a single turn emitting more than 20 blocks (a wide parallel tool fan-out) can still miss. This uses 2 of the 4 breakpoints allowed per request, so there is room for intermediate markers. Filed as #205.

**Not a regression, but worth knowing.** The tail breakpoint writes a cache entry on essentially every request, so the last turn of a conversation pays a ~1.25x write on its delta that nothing ever reads. That was already true for text-ending turns before this change, and it is dwarfed by what the fix recovers — visible above as the 638-token per-turn write against a 4153-token read.

## References

- #205 — the 20-block cache lookback, split out rather than folded in here.
- Follow-up, deliberately not in this PR: cloudv2's `ai-agent` needs an ai-sdk-go release and a `go.mod` bump to pick this up. That bump is a separate change and should not gate this one.

### DIFF
--- a/providers/anthropic/cache_conversation_test.go
+++ b/providers/anthropic/cache_conversation_test.go
@@ -1,0 +1,132 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package anthropic_test
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/redpanda-data/ai-sdk-go/llm"
+	"github.com/redpanda-data/ai-sdk-go/providers/anthropic"
+	"github.com/redpanda-data/ai-sdk-go/providers/anthropic/anthropictest"
+	"github.com/redpanda-data/ai-sdk-go/providers/testutil"
+)
+
+// TestAnthropicConversationCaching_Integration is the end-to-end proof that an
+// agentic conversation caches, against the live API.
+//
+// It replays the shape of a real tool-calling loop: every turn appends an
+// assistant tool_use and a user tool_result, so from turn 2 on the last message
+// carries no text block at all. Two things are being established that a
+// request-shape unit test cannot:
+//
+//  1. Anthropic accepts cache_control on a tool_result block. A rejected
+//     breakpoint is a 400 on the whole request, so any Generate error fails here.
+//  2. The conversation, not just the static tools+system prefix, is cached.
+//     The signature is cache_read GROWING turn over turn. When the breakpoint is
+//     restricted to text blocks the tool_result turns go unmarked, no cache entry
+//     is ever written past the system blocks, and cache_read sits FLAT at the
+//     system-prefix size while input_tokens climbs with the history.
+func TestAnthropicConversationCaching_Integration(t *testing.T) {
+	t.Parallel()
+
+	apiKey := anthropictest.GetAPIKeyOrSkipTest(t)
+
+	provider, err := anthropic.NewProvider(apiKey, anthropic.WithTimeout(3*time.Minute))
+	require.NoError(t, err)
+
+	// Keep the output budget small: only usage accounting matters here.
+	model, err := provider.NewModel(anthropictest.TestModelName, anthropic.WithMaxTokens(64))
+	require.NoError(t, err)
+
+	tools := []llm.ToolDefinition{{
+		Name:        "lookup_record",
+		Description: "Look up a record by id.",
+		Parameters:  json.RawMessage(`{"type":"object","properties":{"id":{"type":"string"}},"required":["id"]}`),
+	}}
+
+	// The cacheable prefix has a per-model token minimum (1024 on Sonnet), so
+	// the system prompt has to clear it on its own for turn 1 to write anything.
+	messages := []llm.Message{
+		{
+			Role:    llm.RoleSystem,
+			Content: []llm.Part{llm.NewTextPart(testutil.GenerateLargePrompt(1800))},
+		},
+		{
+			Role:    llm.RoleUser,
+			Content: []llm.Part{llm.NewTextPart("Look up record alpha.")},
+		},
+	}
+
+	ctx := context.Background()
+
+	const turns = 4
+
+	reads := make([]int, 0, turns)
+	inputs := make([]int, 0, turns)
+
+	for turn := 1; turn <= turns; turn++ {
+		resp, err := model.Generate(ctx, &llm.Request{Messages: messages, Tools: tools})
+		require.NoError(t, err, "turn %d: a rejected cache_control breakpoint surfaces as a 400 here", turn)
+		require.NotNil(t, resp.Usage)
+
+		reads = append(reads, resp.Usage.CachedInputTokens)
+		inputs = append(inputs, resp.Usage.InputTokens)
+
+		t.Logf("turn %d: input %d, cache_read %d, cache_write %d",
+			turn, resp.Usage.InputTokens, resp.Usage.CachedInputTokens,
+			resp.Usage.CacheCreation5mTokens+resp.Usage.CacheCreationUnknownTTLTokens)
+
+		// Append a fixed tool round-trip rather than the model's own reply: the
+		// cache is a byte-prefix match, so the history has to be reproduced
+		// exactly on the next turn. A non-deterministic assistant turn would
+		// invalidate the prefix and make this test measure nothing.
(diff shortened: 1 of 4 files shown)


birdayz:
@claude review this PR

birdayz:
## Round 1 summary (commits a62919e1..98fa8ff7)

### All items addressed

| Round | Item: Fix | Commit |
|-------|-----------|--------|
| 1 | Drift risk: `default: continue` silently drops the breakpoint if image/document support lands later. Took the "handle them now" option — `OfImage` and `OfDocument` arms added, and the doc comment now states why unreachable types are covered. | requested by @claude [`98fa8ff7`](https://github.com/redpanda-data/ai-sdk-go/commit/98fa8ff7) |
| 1 | Thinking-only last message untested: added `TestCacheBreakpointOnThinkingOnlyTurn`, asserting the turn goes unmarked *and* the system breakpoint survives. | requested by @claude [`98fa8ff7`](https://github.com/redpanda-data/ai-sdk-go/commit/98fa8ff7) |
| 1 | Unguarded dereferences at `cache_tool_result_test.go:97` and `:128`: `require.NotNil` added, matching line 61. | requested by @claude [`98fa8ff7`](https://github.com/redpanda-data/ai-sdk-go/commit/98fa8ff7) |
| 1 | Doc comment read as an off-by-one ("three types... those four"): rewritten to state the rule rather than a count. | requested by @claude [`98fa8ff7`](https://github.com/redpanda-data/ai-sdk-go/commit/98fa8ff7) |
| 1 | `OfToolUse` arm is unreachable and its acceptance unverifiable: agreed. PR description now says so explicitly — the arm is defensive, resting on the SDK field inventory, not on a live-API observation. Arm kept; it costs nothing and is correct if ever reached. | requested by @claude (description only) |
| 1 | 20-block lookback should not get lost: filed as #205 and linked from the description. | requested by @claude [#205](https://github.com/redpanda-data/ai-sdk-go/issues/205) |
| 1 | Cost note on the tail write: not a regression, no code change. Recorded in the description with the measured numbers (638-token write against a 4153-token read). | requested by @claude (description only) |

### Two things you flagged as unverified, now verified

You noted the sandbox blocked `go test` and the module cache. Both are settled:

**The `CacheControl` field inventory, read from the pinned SDK.** Of the 19 `Of*` variants in `BetaContentBlockParamUnion` (`anthropic-sdk-go@v1.27.1/betamessage.go:2555`), exactly two lack a `CacheControl` field: `BetaThinkingBlockParam` and `BetaRedactedThinkingBlockParam`. Every other variant — including `BetaToolResultBlockParam`, `BetaImageBlockParam` and `BetaRequestDocumentBlockParam` — has one. So the switch matches the SDK, not just the docs.

**The live-API growth assertion passed in CI**, which is the claim the shape tests cannot make. From the `Test` job on a62919e1:

```
turn 1: input 3, cache_read    0, cache_write 2878
turn 2: input 5, cache_read 2878, cache_write  637
turn 3: input 5, cache_read 3515, cache_write  638
turn 4: input 5, cache_read 4153, cache_write  638
```

`cache_read` grows 2878 -> 3515 -> 4153 across turns whose last message is a `tool_result`-only user turn. That establishes both halves at once: Anthropic accepted the `tool_result` breakpoint (a rejection would have been a 400 on `Generate`), and the conversation is genuinely being cached rather than the marker merely being set. On the old mapper this would have read a flat 2878.

### Note on the second reviewer

The independent codex review could not run this round — ChatGPT credits exhausted and the Bedrock fallback's SSO session expired. No findings from it either way; not treating that as a clean bill of health.

All previous threads and comments minimized. Requesting fresh review.

birdayz:
@claude looks good now?

birdayz:
## Round 2 summary (commit dca56d20)

| Round | Item: Fix | Commit |
|-------|-----------|--------|
| 2 | `lastMessageHasCacheMarker` narrower than the mapper — `assert.False` sites at `cache_default_test.go:138` and `cache_tool_result_test.go:162` could pass silently on a correctly-marked image/document turn: `OfImage` and `OfDocument` cases added, comment now states the direction-of-failure reason. | requested by @claude [`dca56d20`](https://github.com/redpanda-data/ai-sdk-go/commit/dca56d20) |

Agreed on the framing note about unreachable shapes: `TestCacheBreakpointOnThinkingOnlyTurn` and the `OfToolUse` arm pin mapper behavior for shapes the API would reject on other grounds. Neither is evidence about API behavior, and the PR description says so.

### CI, including the check you flagged as still running

All four green on `dca56d20`. The **Test** job — the one that runs the live-API growth assertion — passed in 2m55s, and this run happened to land inside the previous run's 5-minute cache TTL, so it is an even cleaner demonstration than the first:

```
turn 1: input 3, cache_read 2878, cache_write 0
turn 2: input 5, cache_read 3515, cache_write 0
turn 3: input 5, cache_read 4153, cache_write 0
turn 4: input 5, cache_read 4791, cache_write 0
```

Every turn served entirely from cache, `cache_read` growing 2878 -> 4791 while `input` stays at 3-5 tokens, on turns whose last message is a `tool_result`-only user turn. No writes at all because the prefix was already resident. On the old mapper `cache_read` would have been pinned at 2878 with the history re-billed as fresh input.

@claude anything left?

malinskibeniamin (approved):

PR #208, opened 2026-08-25
feat(llmagent): deterministic context compaction

weeco:
## Motivation

Long tool-heavy sessions outgrow the model's context window and die with a provider 400. Nothing in the SDK reduced the history, so an agent doing real research work had a hard lifetime limit.

## What this adds

Opt-in via `llmagent.WithCompaction`, off by default. When the estimated request crosses 80% of the usable window, the runtime rewrites the session toward 60%, cheapest damage first:

- **Prune**: already-read tool results become small JSON markers (tool name, error status, a 200-char preview, "re-run if needed"). The newest 5 results stay verbatim.
- **Drop**: if pruning is not enough, the oldest turns are removed whole. The newest user message and everything after the last assistant message are never touched, so the model never loses unread content.
- **Burst division**: parallel tool results are capped at collection time, so a single turn cannot assemble a request that can never fit.
- **Calibration**: provider-reported usage corrects the char-based token estimate per session (dense content like fetched web pages tokenizes well below the heuristic). Persists in session metadata; apps can seed it with `CalibrateSession`.
- **Failure behavior**: a provider overflow forces one strictly smaller retry; the terminal failure is a typed error wrapping `llm.ErrContextOverflow` with actionable numbers. Each pass emits a `StatusStageCompaction` event.

No LLM call involved: the rewrite is deterministic, idempotent, and keeps the prompt prefix byte-identical between compactions so caching keeps working.

Screenshot 2026-08-25 at 13-19-03


agent/llmagent/compaction.go
```diff
@@ -0,0 +1,389 @@
...
+	// a tool_result whose tool_use was dropped wedges the session permanently.
+	//
+	// The cut may reach limit itself: every limit candidate is legal by
+	// construction. The newest user message carries no tool responses by
+	// definition, a tail-floor turn starts at an assistant message, and
+	// frontier-1 is only a bound when the frontier carries tool results -
+	// then it is the assistant message that issued them and dropping it
+	// would orphan them. A frontier of plain text protects nothing before
+	// it: the assistant answer preceding a fresh user message is droppable.
+	limit := min(newestUserIndex(msgs), tailFloorStart(msgs, minTail))
```
MaciejWrzesniewski-Redpanda, line 281: The drop step never checks the role of the message it lands on, so what is left can start with an assistant message. Three messages are enough:

```
0  user       "<long question>"   <- most of the history, dropping it alone hits the target
1  assistant  "<answer>"          <- cut stops here: no ToolResponsePart, role never checked
2  user       "<follow-up>"       <- newestUserIndex = 2

limit = min(newestUserIndex, tailFloorStart(minTail)) = min(2, 3) = 2
msgs[0] is dropped, cut stops at 1
head is now assistant
```

Providers reject a first message that is not user-role, and the cost is more than one failed call. `ensureFits` saves the compacted history at compaction.go:350 before the request goes out, and the 400 is an `invalid_request_error` whose message does not match `isContextOverflowMessage`, so it never becomes `ErrContextOverflow` and `reduceAfterOverflow` never fires. Every later turn sends the same saved history and gets the same 400.

Making the cut skip assistant messages looks like the fix, and it does work above, but not once tool turns are in the way:

```
0  user               "research X"
1  assistant          tool_call        <- tailFloorStart(3) = 1, last 3 turns start here
2  user(tool_result)
3  assistant          tool_call
4  user(tool_result)
5  assistant          tool_call        <- issued the pending result, so it must survive
6  user(tool_result)                   <- frontierStart = 6, not read yet
7  user               "now summarise"  <- newestUserIndex = 7

limit = min(newestUserIndex, tailFloorStart(minTail)) = min(7, 1) = 1
```

The cut may only land in [0, 1]. That single candidate is the assistant at index 1, and the next user message is index 7 - cutting to there would drop the question and every tool turn. So the walk either lands illegally or does not move at all.

What worked: when the retained head is an assistant message, put a user message in front of it.

```
before   [ assistant "<answer>", user "<follow-up>" ]
after    [ user "(earlier turns were removed to fit the context window)",
           assistant "<answer>", user "<follow-up>" ]
```

agent/llmagent/compaction.go
```diff
@@ -0,0 +1,389 @@
...
+		}
+	}
+
+	return 0
+}
+
+// newestUserIndex returns the index of the newest user message that carries
+// no tool responses - the message driving the current task - or len(msgs)
+// when there is none. It is never dropped, in any mode.
+func newestUserIndex(msgs []llm.Message) int {
```
MaciejWrzesniewski-Redpanda, line 135: The newest user message is never dropped, and nothing after it can be dropped either. In a single invocation that message is the question at index 0, so nothing in the history can be dropped and the run fails once the context window fills. This happens two ways, and each needs a different fix.

Case 1: massive initial prompt. Compaction cannot fix this, since the prompt is the thing being protected. IMHO better to reject it up front than run tools and fail later.

Case 2: long chain of turns triggering compaction before the invocation closes. Seems like the LLM summariser case, plus cutting the tool results.

agent/llmagent/compaction.go
weeco: Good find, thanks - fixed

agent/llmagent/compaction.go
weeco: Yeah I think that's fine to ignore for now. 

**Re 1:** If either system prompt or initial prompt by itself is already over the hard limit it will already fail before the provider call. Whether or not it requests tool I can't know upfront.

**Re 2:** Agree, LLM summarizer case, leaving that for you :).

MaciejWrzesniewski-Redpanda: Thanks, ack, all noted 👍

MaciejWrzesniewski-Redpanda (approved):

PR #212, opened 2026-09-02
fix(bedrock): flatten stale tool blocks when toolConfig is absent

birdayz:
## What

Flatten `toolUse`/`toolResult` content blocks to plain text in the Bedrock request mapper when the outgoing request has no `toolConfig` to cover them, instead of letting Bedrock reject the whole call.

## Why

Bedrock's Converse API requires `toolConfig` on a request whenever the message history contains `toolUse`/`toolResult` blocks, regardless of whether the current turn has any tools to offer. Remove a tool from an agent and every later turn on a session that used it hits:

```
ValidationException: The toolConfig field must be defined when using toolUse and toolResult content blocks.
```

This is a documented, retroactive constraint specific to Bedrock's Converse API — Anthropic's and OpenAI's own APIs don't enforce it, so nothing here reuses an existing pattern in this codebase. It's also a widely-reported problem: langchain-aws, langgraphjs, the Vercel AI SDK, opencode, Dify and Roo-Code have all hit the identical error string. langchain-aws is the only one with a merged fix.

## Implementation details

`buildConverseParams` in `providers/bedrock/request_mapper.go` already builds `ToolConfig` conditionally on `len(req.Tools) > 0` for the current turn, but `mapAssistantMessage`/`mapUserMessage` unconditionally re-serialize historical tool blocks. Added a check after tool mapping: if `p.tools` ends up nil and the mapped history still contains tool blocks, run `convertToolBlocksToText` over the messages.

Ported langchain-aws's fix (PR #595) verbatim in spirit: `toolUse` becomes `"[Called {name} with parameters: {json}]"` (or `"[Called {name}]"` with no parameters), `toolResult` becomes `"[Tool output: {content}]"` (dropped entirely if it carries no representable content). Everything else passes through untouched. No fake `toolConfig` is synthesized — Bedrock's `ToolConfiguration.tools` isn't documented to accept an empty list, so that route wasn't viable.

Placed as a single check in the shared `buildConverseParams`, covering both `ToConverseInput` and `ToConverseStreamInput`, rather than duplicating the check at each call site the way the Python fix does.

Three pre-existing unit tests (`TestRequestMapper_ToolResponse`, `TestRequestMapper_ToolResponseError`, `TestRequestMapper_AssistantWithToolUse`) built tool-bearing requests without ever setting `Tools`, which is now a legitimate trigger for flattening. Added a matching `Tools` entry to each so they keep exercising the tools-bound path they were meant to test.

## References

https://github.com/langchain-ai/langchain-aws/pull/595
https://github.com/langchain-ai/langchain-aws/issues/591

### DIFF
--- a/providers/bedrock/provider_test.go
+++ b/providers/bedrock/provider_test.go
@@ -872,6 +872,9 @@ func TestRequestMapper_ToolResponse(t *testing.T) {
 				},
 			),
 		},
+		Tools: []llm.ToolDefinition{
+			{Name: "search", Description: "Search", Parameters: json.RawMessage(`{"type":"object"}`)},
+		},
 	}
 
 	input, err := mapper.ToConverseInput(req)
@@ -906,6 +909,9 @@ func TestRequestMapper_ToolResponseError(t *testing.T) {
 				},
 			),
 		},
+		Tools: []llm.ToolDefinition{
+			{Name: "search", Description: "Search", Parameters: json.RawMessage(`{"type":"object"}`)},
+		},
 	}
 
 	input, err := mapper.ToConverseInput(req)
@@ -936,6 +942,9 @@ func TestRequestMapper_AssistantWithToolUse(t *testing.T) {
 					Arguments: json.RawMessage(`{"query":"cats"}`),
 				},
 			),
+		},
+		Tools: []llm.ToolDefinition{
+			{Name: "search", Description: "Search", Parameters: json.RawMessage(`{"type":"object"}`)},
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


birdayz:
@claude review this PR

birdayz:
## 🤖 Codex review

Automated non-Claude second opinion via `codex review` — tier **chatgpt**. These are an external model's suggestions; triage before acting.

**Findings:** P1=0 · P2=0 · P3=0

The change consistently flattens stale Bedrock tool blocks only when tool configuration is absent, while preserving structured blocks when tools remain available. The Bedrock tests and new lint checks pass.

Script done on 2026-09-02 22:58:10+02:00 [COMMAND_EXIT_CODE="0"]

birdayz:
Fixed both in 7ba00026:

- #1 (real bug): toolResultText no longer drops an empty result. Added the regression through ToConverseInput with caching on, per your repro.
- #2: Status is now rendered as "[Tool error: ...]" instead of merging into "[Tool output: ...]".

Left #3's smaller notes as-is except the toolConfig-reconstruction rationale, which is now a comment at the call site. The non-text/JSON skip and no-separator points stay unreachable given mapToolResultBlock only ever emits one Text block, so didn't add speculative handling for that.

PR #217, opened 2026-09-06
feat(openai): add gpt-6 astra

malinskibeniamin:
## Summary
- Register `openai.ModelGPT6Astra` / `catalog.ModelGPT6Astra` with official limits, modalities, release date and cutoff.
- Accept low/medium/high/xhigh/max reasoning; reject none/minimal and unsupported temperature/top-p through existing validation.
- Encode Standard input/output/cache-read/cache-write pricing on both sides of the 272K threshold.
- Regenerate catalog snapshot; GPT-5.6 Sol's derived successor becomes Astra without changing its authored entry.

## Sources
- https://developers.openai.com/api/docs/models/gpt-6-astra
- https://developers.openai.com/api/docs/guides/latest-model
- https://openai.com/index/safety-overview-gpt-6-astra/

## Scope
OpenAI Responses provider registration only. No dependency changes, Bedrock offering, new async/steering APIs, or Fast/Batch/Flex pricing modes. Vendor access is subject to rollout.

## Verification
- RED: new catalog/pricing tests failed for missing Astra registration.
- PASS: `go test -short ./providers/openai ./catalog/... ./pricing/...`.
- PASS: `task catalog:check`, `task license:check`, `git diff --check`.
- PASS: golangci-lint with `--new-from-rev=origin/main`: 0 new issues.
- Full `task lint`: 35 existing issues outside this change.
- Full `task test:unit`: Docker-backed schema-registry tests fail because Docker is unavailable; focused packages pass.
- Go repository; Bun lint/type-check commands do not apply.

## Dogfood evidence
- PASS: standalone consumer calls the public provider/catalog/pricing APIs; creates Astra at max effort, rejects none/temperature/top-p, and calculates both pricing brackets.
- Limits: no live inference or 1M-context probe; no OpenAI API key configured. No transport code changed.

### DIFF
--- a/catalog/facts_data.go
+++ b/catalog/facts_data.go
@@ -36,6 +36,7 @@ const (
 
 	// OpenAI models.
 
+	ModelGPT6Astra     ModelID = "openai/gpt-6-astra"
 	ModelGPT5          ModelID = "openai/gpt-5"
 	ModelGPT5Mini      ModelID = "openai/gpt-5-mini"
 	ModelGPT5Nano      ModelID = "openai/gpt-5-nano"
@@ -228,6 +229,13 @@ var defaultRegistry = Registry{
 		DisplayName: "GPT-5.5", Series: "gpt",
 		Released: MustDate("2026-04-23"), Knowledge: MustDate("2025-12-01"),
 		Description: "GPT-5.5 is OpenAI’s frontier model designed for complex professional workloads, building on GPT-5.4 with stronger reasoning, higher reliability, and improved token efficiency on hard tasks.",
+	},
+	// Announcement: https://openai.com/index/gpt-6-astra/
+	// Cutoff: https://developers.openai.com/api/docs/models/gpt-6-astra
+	ModelGPT6Astra: {
+		DisplayName: "GPT-6 Astra", Series: "gpt",
+		Released: MustDate("2026-09-03"), Knowledge: MustDate("2026-04-30"),
+		Description: "GPT-6 Astra is OpenAI's flagship model for complex reasoning, coding, and long-running professional work.",
 	},
 	ModelGPT5_6Sol: {
 		DisplayName: "GPT-5.6 Sol", Series: "gpt",
--- a/catalog/facts_data.go
+++ b/catalog/facts_data.go
@@ -245,14 +245,14 @@ var defaultRegistry = Registry{
 		Released: MustDate("2026-07-09"), Knowledge: MustDate("2026-02-16"),
 		Description: "GPT-5.6 Luna is a fast, cost-efficient model in OpenAI's GPT-5.6 series.",
 	},
-
 	// Announcement: https://openai.com/index/gpt-6-astra/
 	// Cutoff: https://developers.openai.com/api/docs/models/gpt-6-astra
 	ModelGPT6Astra: {
 		DisplayName: "GPT-6 Astra", Series: "gpt",
 		Released: MustDate("2026-09-03"), Knowledge: MustDate("2026-04-30"),
 		Description: "GPT-6 Astra is OpenAI's flagship model for complex reasoning, coding, and long-running professional work.",
-	}, ModelGPT4o: {
+	},
+	ModelGPT4o: {
 		DisplayName: "GPT-4o", Series: "gpt",
 		Released: MustDate("2024-05-13"), Knowledge: MustDate("2023-09-30"),
 		Description: "GPT-4o (\"o\" for \"omni\") is OpenAI's latest AI model, supporting both text and image inputs with text outputs.",
--- a/catalog/facts_data.go
+++ b/catalog/facts_data.go
@@ -36,7 +36,6 @@ const (
 
 	// OpenAI models.
 
-	ModelGPT6Astra     ModelID = "openai/gpt-6-astra"
 	ModelGPT5          ModelID = "openai/gpt-5"
 	ModelGPT5Mini      ModelID = "openai/gpt-5-mini"
 	ModelGPT5Nano      ModelID = "openai/gpt-5-nano"
@@ -52,6 +51,7 @@ const (
 	ModelGPT5_6Sol     ModelID = "openai/gpt-5.6-sol"
 	ModelGPT5_6Terra   ModelID = "openai/gpt-5.6-terra"
 	ModelGPT5_6Luna    ModelID = "openai/gpt-5.6-luna"
+	ModelGPT6Astra     ModelID = "openai/gpt-6-astra"
 	ModelGPT4o         ModelID = "openai/gpt-4o"
 	ModelGPT4oMini     ModelID = "openai/gpt-4o-mini"
 	ModelGPT4Turbo     ModelID = "openai/gpt-4-turbo"
@@ -168,7 +168,7 @@ var defaultRegistry = Registry{
 	// ---- OpenAI ----
 	//
 	// The flagship line runs gpt-3.5-turbo → gpt-4-turbo → gpt-4o →
-	// gpt-4.1 → gpt-5 → 5.1 → 5.2 → 5.4 → 5.5 → gpt-5.6-sol; the -mini
+	// gpt-4.1 → gpt-5 → 5.1 → 5.2 → 5.4 → 5.5 → gpt-5.6-sol → gpt-6-astra; the -mini
 	// ladder ends in terra, the -nano ladder in luna. The chat-tuned
 	// "instant" models and the pro models are their own lines.
 	ModelGPT5: {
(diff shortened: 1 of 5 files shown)


catalog/facts_data.go
```diff
@@ -229,6 +230,13 @@ var defaultRegistry = Registry{
 		Released: MustDate("2026-04-23"), Knowledge: MustDate("2025-12-01"),
 		Description: "GPT-5.5 is OpenAI’s frontier model designed for complex professional workloads, building on GPT-5.4 with stronger reasoning, higher reliability, and improved token efficiency on hard tasks.",
 	},
+	// Announcement: https://openai.com/index/gpt-6-astra/
+	// Cutoff: https://developers.openai.com/api/docs/models/gpt-6-astra
+	ModelGPT6Astra: {
+		DisplayName: "GPT-6 Astra", Series: "gpt",
```
malinskibeniamin: **Priority: P3**

**Series lineage comment is now stale.** Adding GPT-6 Astra to `Series: "gpt"` with `Released: 2026-09-03` makes it the newest member of the flagship line, so `Catalog.Successor("openai/gpt-5.6-sol")` now returns astra — visible in the snapshot as the new `"replacement": "openai/gpt-6-astra"` on the sol offering. The registry's own documentation at `catalog/facts_data.go:170-173` still states the line "runs gpt-3.5-turbo → … → gpt-5.6-sol", which now contradicts the data it is documenting.

This file explicitly treats series discipline as maintained prose (`// Series discipline: a Series is a NON-BRANCHING succession line`), so the comment is the reference readers use to reason about replacement math.

Correction: extend the flagship-line comment to `… → 5.5 → gpt-5.6-sol → gpt-6-astra`, and note the -mini/-nano ladders still end at terra/luna.

Verify: read `catalog/facts_data.go:168-173` and confirm the described tail matches the newest `Series: "gpt"` entry by `Released`.

*Automated `/review`.*

catalog/facts_data.go
```diff
@@ -36,6 +36,7 @@ const (
 
 	// OpenAI models.
 
+	ModelGPT6Astra     ModelID = "openai/gpt-6-astra"
```
malinskibeniamin: **Priority: P3**

**New constant breaks the block's generation ordering.** The OpenAI constant block runs oldest-to-newest within the GPT-5 line (`ModelGPT5` → `ModelGPT5_1` → … → `ModelGPT5_5` → `ModelGPT5_6Sol` → `ModelGPT5_6Terra` → `ModelGPT5_6Luna`) before the legacy 4o/4-turbo tail; `ModelGPT6Astra` is inserted ahead of `ModelGPT5`. The registry insertion has the same issue: the astra entry lands between `ModelGPT5_5` (2026-04-23) and `ModelGPT5_6Sol` (2026-07-09) rather than after the 5.6 group.

No behavioral impact — lookups are by key and successor math is date-driven — but it makes the next generation's diff harder to place consistently.

Correction: move the constant to follow `ModelGPT5_6Luna` (facts_data.go:54) and the registry entry to follow the `ModelGPT5_6Luna` fact, keeping one convention.

Verify: scan the const block top-to-bottom and confirm GPT identifiers ascend by generation.

*Automated `/review`.*

providers/openai/constants.go
```diff
@@ -63,6 +63,9 @@ const (
 	// stays priceable.
 	ModelGPT5_3ChatLatest = shared.ChatModelGPT5_3ChatLatest
 
+	// ModelGPT6Astra is OpenAI's GPT-6 flagship reasoning model.
+	ModelGPT6Astra = "gpt-6-astra"
```
malinskibeniamin, line 68: **Priority: P3**

**Raw-string constant omits the rationale the file uses elsewhere.** Every other current-generation entry aliases the vendor SDK constant (`shared.ChatModelGPT5_6Sol`, `shared.ChatModelGPT5_4`, …). The one existing raw string documents why: `ModelGPT5_5 = "gpt-5.5"` carries `// Raw string until the OpenAI SDK adds the constant.` (constants.go:78-80). The new `ModelGPT6Astra = "gpt-6-astra"` has no such note, so a future reader cannot tell whether the literal is a deliberate stopgap or an oversight to normalize.

Correction: append the same one-line rationale, for example `// Raw string until the OpenAI SDK adds the constant.`

Verify: compare the doc comments on `ModelGPT5_5` and `ModelGPT6Astra`.

*Automated `/review`.*

providers/openai/astra_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/redpanda-data/ai-sdk-go/catalog"
+	"github.com/redpanda-data/ai-sdk-go/llm"
+	"github.com/redpanda-data/ai-sdk-go/pricing"
+)
+
+func TestGPT6AstraCatalog(t *testing.T) {
+	t.Parallel()
+
+	offering, ok := Catalog().Lookup("gpt-6-astra")
```
malinskibeniamin: **Priority: P3**

**Test hardcodes the model ID the PR just added a constant for.** `Catalog().Lookup("gpt-6-astra")` here, plus `provider.NewModel("gpt-6-astra", …)` at lines 48 and 54, use string literals while the same commit introduces `ModelGPT6Astra` at `providers/openai/constants.go:67` whose documented purpose is to "avoid typos". Sibling tests use the constants (`openai_test.go:176-178` uses `ModelGPT5_6Luna`/`Terra`/`Sol`). A future rename of the offering ID would leave these literals silently pointing at a missing entry, failing on `require.True(t, ok)` rather than at compile time.

Correction: replace the four literals with `ModelGPT6Astra` (keeping the `catalog.ModelID("openai/gpt-6-astra")` assertion, which intentionally pins the wire value).

Verify: `grep -n '"gpt-6-astra"' providers/openai/astra_test.go` returns only the catalog.ModelID assertion.

*Automated `/review`.*

providers/openai/astra_test.go
```diff
@@ -0,0 +1,87 @@
...
+	assert.Equal(t, []catalog.Modality{catalog.ModalityText}, offering.Modalities.Output)
+	assert.True(t, offering.Capabilities.Tools)
+	assert.True(t, offering.Capabilities.StructuredOutput)
+	assert.True(t, offering.Capabilities.Streaming)
+
+	provider, err := NewProvider("test-key")
+	require.NoError(t, err)
+
+	for _, effort := range offering.Reasoning.Efforts {
+		model, err := provider.NewModel("gpt-6-astra", WithReasoningEffort(effort))
```
malinskibeniamin: **Priority: P3**

**Effort loop asserts nothing about the effort.** The loop constructs a model per supported effort but only checks `model.Name()`, which is independent of `WithReasoningEffort`. It proves the option is accepted, not that `xhigh`/`max` — the two efforts new to this entry relative to most GPT entries — actually reach the provider request. The established pattern asserts the mapped value: `openai_test.go:196-202` maps a request and checks `assert.Equal(t, shared.ReasoningEffortMax, apiReq.Reasoning.Effort)`.

As written, a regression in effort mapping for astra would keep this test green.

Correction: inside the loop, cast to `*Model`, call `requestMapper.ToProvider(&llm.Request{…})`, and assert `apiReq.Reasoning.Effort` equals the effort under test (and `apiReq.Model` equals the offering ID).

Verify: temporarily drop the reasoning-effort mapping for this model and confirm the test fails.

*Automated `/review`.*

malinskibeniamin (commented):
Automated `/review`: 5 finding(s).

catalog/facts_data.go
malinskibeniamin: Fixed in 7365a82. Extended flagship lineage; Terra/Luna remain separate. Focused race tests and new-issue lint pass.

catalog/facts_data.go
malinskibeniamin: Fixed in 7365a82. Moved constant and fact after the 5.6 group; snapshot unchanged. Focused race tests and new-issue lint pass.

providers/openai/constants.go
malinskibeniamin: Fixed in 7365a82. Documented raw-string stopgap; installed SDK lacks Astra constant. Focused race tests and new-issue lint pass.

providers/openai/astra_test.go
malinskibeniamin: Fixed in 7365a82. Reused the constant, preserving independent canonical wire-ID assertion. Focused race tests and new-issue lint pass.

providers/openai/astra_test.go
malinskibeniamin: Fixed in 7365a82. Asserted request model and effort for each effort. Forcing low effort fails the test; restored mapping passes. Focused race tests and new-issue lint pass.

weeco (approved):

PR #218, opened 2026-09-06
feat(meta): add muse spark 1.3 provider

malinskibeniamin:
## Output-budget policy
Meta documents a shared context window, not a separate numeric output maximum. The catalog records a conservative SDK ceiling of 32,768 for explicitly supplied budgets, labelled `output_token_limit_source=sdk_conservative_limit`. Omitted budgets now remain omitted, allowing the server default. This is not a claimed vendor maximum.

## Summary
- Add a thin `providers/meta` provider for Standard-tier `muse-spark-1.3`, reusing the existing OpenAI Responses transport.
- Preserve Meta provider identity and expose immutable reasoning-effort metadata.
- Register canonical facts, documented modalities, minimal-through-max reasoning, flat pricing, and the generated snapshot.
- Reject unknown/Contributor model IDs, unsupported none reasoning, invalid output budgets, and options the shared transport would silently ignore.
- Document transport limitations; add local HTTP generation/streaming tests and key-gated provider/agent conformance suites.

## Scope and limits
- Direct Meta API, Standard tier only. No Contributor/data-training tier.
- No dependencies or duplicated request/response mapper.
- Reuse `openai.ProviderOption` and `openai.Option` for supported controls.
- Binary inputs, encrypted reasoning replay, built-in Meta tools, and newly wiring sampling controls are outside this draft. Metadata represents vendor modalities; it does not claim mapper coverage.
- Knowledge cutoff left unknown; Meta warns that 1.3 audio understanding quality may be degraded.
- Independent PR against main; no dependency on the Astra PR.

## Sources
- https://dev.meta.ai/docs/models
- https://dev.meta.ai/docs/pricing-rate-limits
- https://dev.meta.ai/docs/reasoning
- https://dev.meta.ai/docs/protocols/responses
- https://dev.meta.ai/docs/api-reference/responses/schemas#create-response
- https://research.meta.ai/blog/introducing-muse-spark-1-3

## Verification
- RED: new tests fail before provider implementation; review regression tests catch unsolicited default budgets and unsupported advertised parameters.
- PASS: `go test -short ./providers/meta ./providers/openai ./catalog/... ./cmd/catalog-snapshot ./pricing/...`.
- PASS: `go test -race -short ./providers/meta ./catalog/... ./cmd/catalog-snapshot`.
- PASS: golangci-lint with `--new-from-rev=origin/main`: 0 new issues.
- PASS: `task catalog:check`, `task license:check`, `git diff --check`.
- Full `task test:unit`: fails only in the two Docker-backed schema-registry packages because Docker is unavailable locally.
- Live integration test skips: `MODEL_API_KEY` is not configured.
- Go repository: Bun frontend lint/type-check commands do not apply.

## Dogfood evidence
- PASS at the local library/HTTP seam: standalone consumer creates the provider/model, sends real HTTP requests to a local fixture endpoint, exercises Generate and GenerateEvents, and observes META_OK from both.
- Break attempts: budgets of 15 and 32,769, none reasoning, unmapped top-p, and canceled context all fail visibly.
- Limits: fixtures prove SDK wiring, not vendor compatibility. No live Meta invocation, maximum-output probe, or large-context probe performed.

## Reviewer guide
Start with `providers/meta/README.md` and known limits, then `provider.go` / `models.go`, tests, and `catalog/snapshot.json`.


providers/meta/models.go
```diff
@@ -0,0 +1,74 @@
...
+			Input:  []catalog.Modality{catalog.ModalityText, catalog.ModalityImage, catalog.ModalityVideo, catalog.ModalityAudio, catalog.ModalityDocument},
+			Output: []catalog.Modality{catalog.ModalityText},
+		},
+		Constraints: llm.ModelConstraints{
+			TemperatureRange: [2]float64{0, 2},
+			MaxInputTokens:   1_048_576,
+			// Meta publishes the shared context window, not a separate output cap.
+			// This is the SDK-enforced bound, not a claim about the vendor limit.
+			MaxOutputTokens:   maxOutputTokens,
+			SupportedParams:   []string{"temperature", "top_p", "max_tokens", "frequency_penalty", "presence_penalty", "reasoning_effort", "reasoning_summary"},
```
malinskibeniamin: **Priority: P2**

**Catalog advertises parameters the provider always rejects.** `SupportedParams` declares `top_p`, `frequency_penalty`, and `presence_penalty`, and `MutuallyExclusive` encodes a `temperature`/`top_p` pair — but the option closure in `providers/meta/provider.go:72` returns an error whenever any of those three is set, so `openai.WithTopP`/`WithFrequencyPenalty`/`WithPresencePenalty` can never succeed for this model. Consumers that build UI or validation from `Model.Constraints()` (`providers/openai/model.go:60` returns exactly these constraints) will surface options that are guaranteed to fail at `NewModel`, and the mutual-exclusion rule is dead code.

Correction: drop `top_p`, `frequency_penalty`, `presence_penalty` (and the `MutuallyExclusive` pair) from `SupportedParams` until the shared Responses mapper serializes them; then remove the wrapper rejection and assert the wire body in `transport_test.go`. Verify: `go test ./providers/meta` plus a constraints assertion that every declared param is accepted by `NewModel`.

*Automated `/review`.*

providers/meta/README.md
```diff
@@ -0,0 +1,63 @@
...
+~~~
+
+Import github.com/redpanda-data/ai-sdk-go/providers/meta and
+github.com/redpanda-data/ai-sdk-go/providers/openai for the shared options.
+openai.WithBaseURL, WithHTTPClient, and WithTimeout configure the
+provider. The default endpoint is https://api.meta.ai/v1.
+
+## Draft limits
+
+- **Output cap requires confirmation before merge.** Meta publishes a
```
malinskibeniamin: **Priority: P2**

**Shipped docs declare an unresolved pre-merge blocker.** The user-facing provider README states "**Output cap requires confirmation before merge.**" and titles the section "Draft limits", yet this is the merged provider documentation. Either the 32,768 cap was confirmed (in which case this text misleads users into thinking the provider is unfinished) or it was not (in which case the PR ships an acknowledged open question as a hard default at `providers/meta/provider.go:63`).

Correction: resolve the cap against `https://dev.meta.ai/docs/models` and reword to a factual "Known limits" section describing the SDK-imposed 32,768 cap and why, without the merge-gate language. Verify: re-read the rendered README and confirm the `output_token_limit_source` attribute wording matches.

*Automated `/review`.*

providers/meta/provider.go
```diff
@@ -0,0 +1,105 @@
...
+// by the shared transport.
+func (p *Provider) NewModel(modelName string, opts ...openai.Option) (llm.Model, error) {
+	offering, ok := Catalog().Lookup(modelName)
+	if !ok {
+		return nil, fmt.Errorf("unsupported Meta model: %s", modelName)
+	}
+
+	options := append(slices.Clone(opts), func(cfg *openai.Config) error {
+		if cfg.MaxTokens == nil {
+			budget := maxOutputTokens
```
malinskibeniamin: **Priority: P3**

**Unconditional default output budget diverges from every other provider.** When the caller supplies no `WithMaxTokens`, this closure injects `max_output_tokens=32768` into every request (`request_mapper.go:67` forwards it), so a Meta request is always capped even though the vendor never published an output maximum. `providers/openai`, `anthropic`, and `google` leave the field unset and let the server choose. Callers who want the server default cannot opt out.

Correction: only enforce the ceiling when a budget is supplied, or keep the default but document/expose a way to omit it. Verify: extend `transport_test.go` with a case asserting the request body when no `WithMaxTokens` is passed, under whichever semantics is chosen.

*Automated `/review`.*

catalog/facts_data.go
```diff
@@ -76,6 +76,9 @@ const (
 	ModelGemini25Flash       ModelID = "google/gemini-2.5-flash"
 	ModelGemini25FlashLite   ModelID = "google/gemini-2.5-flash-lite"
 
+	// ModelMuseSpark13 identifies Meta's Muse Spark 1.3.
+	ModelMuseSpark13 ModelID = "meta/muse-spark-1.3"
```
malinskibeniamin, line 82: **Priority: P3**

**Const and registry entries break the file's vendor grouping.** `ModelMuseSpark13` is inserted at the end of the Google/Gemini const block, immediately above the `// Models from vendors offered only through Bedrock.` divider, so it reads as a Bedrock-only or Google model. The matching `defaultRegistry` entry (line 105) is placed above the `// ---- Anthropic ----` section header rather than in its own `// ---- Meta ----` section.

Correction: add a dedicated Meta const group and a `// ---- Meta ----` registry section between Google and OpenAI, matching the existing convention. Verify: `go test ./catalog/...` (behavior unchanged) and visual scan of the section headers.

*Automated `/review`.*

providers/meta/provider_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/redpanda-data/ai-sdk-go/catalog"
+	"github.com/redpanda-data/ai-sdk-go/llm"
+	"github.com/redpanda-data/ai-sdk-go/providers/openai"
+)
+
+func TestMuseSpark13Catalog(t *testing.T) {
```
malinskibeniamin, line 28: **Priority: P3**

**No conformance fixture for the new provider.** Every other provider ships a `providers/conformance` fixture (`providers/anthropic/anthropic_conformance_test.go`, `bedrock_conformance_test.go`, `google_conformance_test.go`, `openai_conformance_test.go`, `openaicompat/deepseek_conformance_test.go`) plus an agent conformance suite; `providers/meta` adds only catalog, option, httptest transport, and a single live smoke test. The shared cross-provider behavioral contract (tool loops, streaming edge cases, error classification) is therefore unverified for Meta.

Correction: add a `conformance.Fixture` implementation for Meta (key-gated like the DeepSeek fixture) so the shared suite runs. Verify: `go test ./providers/meta -run Conformance` with a key, and `go test -short ./providers/meta` to confirm it skips cleanly.

*Automated `/review`.*

malinskibeniamin (commented):
Automated `/review`: 6 finding(s).

providers/meta/README.md
```diff
@@ -0,0 +1,63 @@
+# Meta Model API
```
malinskibeniamin, line 1: **Priority: P3**

**Root README not updated for the new provider.** `README.md:6` enumerates supported providers ("OpenAI, Anthropic, Google Gemini, AWS Bedrock, and any OpenAI-compatible API") and gives a per-provider quickstart block for each; Meta is absent, so the provider is undiscoverable from the SDK entrypoint doc even though it is registered in the snapshot (`cmd/catalog-snapshot/main.go:55`).

Correction: add Meta to the provider list and a short quickstart block in the root `README.md` mirroring the Google/Anthropic sections. Verify: grep the root README for `providers/meta` and confirm the snippet compiles against `meta.NewProvider`/`meta.ModelMuseSpark13`.

*Automated `/review`.*

providers/meta/models.go
malinskibeniamin: Fixed in 11bef2d. Removed unsupported parameters and dead mutual-exclusion metadata; added declared-parameter acceptance regression test and regenerated snapshot. Focused race tests and new-issue lint pass.

providers/meta/README.md
malinskibeniamin: Fixed in 11bef2d. Replaced draft-gate prose with known limits. 32,768 is explicitly an SDK ceiling on supplied budgets, not a vendor maximum. Focused race tests and new-issue lint pass.

providers/meta/provider.go
malinskibeniamin: Fixed in 11bef2d. Omitted budgets now remain omitted. HTTP tests cover omission, 1,024 and 32,768 in generation and streaming; regression verified RED to GREEN. Focused race tests and new-issue lint pass.

catalog/facts_data.go
malinskibeniamin: Fixed in 11bef2d. Added dedicated Meta groups with correctly attached source documentation. Focused race tests and new-issue lint pass.

providers/meta/provider_test.go
malinskibeniamin: Fixed in 11bef2d. Added provider and agent conformance suites. Short tests pass; non-short suite skips without MODEL_API_KEY. Live tests not run: no key configured. Focused race tests and new-issue lint pass.

providers/meta/README.md
malinskibeniamin: Fixed in 11bef2d. Added root provider listing and quickstart; extracted snippet compiles. Focused race tests and new-issue lint pass.

ZK-RPD (approved):
LGTM

MikolajGagatekRP (approved):

PR #221, opened 2026-09-08
feat(vertex): add the Google Vertex AI model catalog

MaciejWrzesniewski-Redpanda:
## Summary

- Adds `providers/vertex`, the Google Vertex AI model catalog: the day-one Gemini and Claude models, their rates, and which model is served at which location. This is [RFC-0014](https://github.com/redpanda-data/cloudv2/blob/main/apps/aigw/docs/rfcs/0014-vertex-ai-provider.md) [milestone M1](https://github.com/redpanda-data/cloudv2/blob/main/apps/aigw/docs/rfcs/0014-vertex-ai-provider.md#-milestones), the catalog half that every later Vertex milestone builds on.
- Data only. It sends no request to Vertex and mints no token. The gateway prices Vertex traffic from this catalog, so Vertex spend prices at $0 until this package is published and pinned.

## Scope

Public surface of the new package:

```
vertex/  (catalog half, M1)
├─ provider.go   Provider -> catalog.Provider (Name, Catalog)
├─ models.go     Catalog(), OfferingForModel(), Model*/ModelMetadata* consts,
│                geminiFlashPricing / claudeSonnet5Pricing / claudeHaiku45Pricing
└─ locations.go  IsModelAvailableAtLocation(), LocationsForModel()
```

Also wires the Vertex catalog into the snapshot generator (`cmd/catalog-snapshot`) and regenerates `catalog/snapshot.json`, the checked-in copy a test keeps in sync.

## Out of Scope

- **Runtime request transport and credential loading.** Routing Gemini and Claude requests through the gateway is [AI-2027](https://redpandadata.atlassian.net/browse/AI-2027), GCP credential resolution is [AI-2019](https://redpandadata.atlassian.net/browse/AI-2019), managed-agent request building is [AI-2031](https://redpandadata.atlassian.net/browse/AI-2031).

## Acceptance Criteria

From [AI-2016](https://redpandadata.atlassian.net/browse/AI-2016):

1. `providers/vertex/` exists on `origin/main` and `Catalog().All()` returns the RFC-0014 day-one catalog of exactly three models (`gemini-3.6-flash`, `claude-haiku-4-5`, `claude-sonnet-5`). The remaining Gemini and Claude models are tracked in [AI-2108](https://redpandadata.atlassian.net/browse/AI-2108).
2. `Catalog().PricingByID()` returns a rate for every offering, with the non-global Gemini rate expressed as a `Selector.Region` override rather than a separate model entry.
3. The location-availability helper answers, for a model and a location, whether Google publishes that pair, and carries the matrix source URL and the transcription date.
4. The conformance and integration suites pass in CI on `origin/main`.
5. A pseudo-version is published and its exact version string is recorded in a comment on this issue.
6. Every rate in the catalog was compared against Google's published Vertex rates on the day the PR merged, and the comparison is recorded in a comment on this issue.

### The `vertex.` key prefix

On Vertex a model keeps its publisher's bare ID, so `claude-sonnet-5` is byte-for-byte the same string the Anthropic-direct catalog already uses. A merged pricing catalog needs unique keys, so each Vertex offering registers under a namespaced key, `vertex.claude-sonnet-5`.

The bare ID (`claude-sonnet-5`) and the publisher (`anthropic`) are not thrown away: they are stored on the offering as the `ModelMetadata*` attributes, so a caller can still find and label the model by its bare ID. They are deliberately not registered as a second catalog key. A `claude-sonnet-5` alias entry would collide with the Anthropic-direct entry all over again, which is the exact collision the prefix exists to avoid.

```mermaid
flowchart TD
  ANT["anthropic-direct offering"] --> kA["catalog key<br/>claude-sonnet-5"]
  VTX["vertex offering<br/>bare model: claude-sonnet-5"] -->|"prefix with vertex."| kV["catalog key<br/>vertex.claude-sonnet-5"]
  VTX -.->|"if keyed bare"| kBad["catalog key<br/>claude-sonnet-5"]
  VTX --> MD["bare ID and publisher kept as<br/>ModelMetadata* attributes, not a key"]

  kA --> M{"merged pricing catalog<br/>every key must be unique"}
  kV --> M
  kBad -. "duplicate key, rejected" .-> M

  classDef bad stroke:#e5534b,color:#e5534b,stroke-width:2px;
  class kBad bad;
```

### Model catalog and pricing

Rates are USD per million tokens, read from Google's [Agent Platform pricing page](https://cloud.google.com/gemini-enterprise-agent-platform/generative-ai/pricing) (its region tabs) on 2026-09-08. Every non-global rate is exactly global x 1.10, flat across the page's =< 200K and > 200K input tiers.

| Model | Scope | Input | Output | Cache read | Cache-creation (5m / 1h) |
|---|---|---|---|---|---|
| gemini-3.6-flash | global | 0.75 | 3.75 | 0.075 | - |
| | us, eu | 0.825 | 4.125 | 0.0825 | - |
| claude-sonnet-5 | global | 2.00 | 10.00 | 0.20 | 2.50 / 4.00 |
| | us, eu, asia-southeast1 | 2.20 | 11.00 | 0.22 | 2.75 / 4.40 |
| claude-haiku-4-5 | global | 1.00 | 5.00 | 0.10 | 1.25 / 2.00 |
| | us-east5, europe-west1, asia-east1 | 1.10 | 5.50 | 0.11 | 1.375 / 2.20 |

**Where each model is served**, read from Google's [locations page](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/locations) on 2026-09-08 (the regions where Google publishes the model and bills it per token on the standard on-demand plan):

| Model | global | us | eu | us-east5 | europe-west1 | asia-southeast1 | asia-east1 |
|---|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| gemini-3.6-flash | ✅ | ✅ | ✅ | | | | |
| claude-sonnet-5 | ✅ | ✅ | ✅ | | | ✅ | |
| claude-haiku-4-5 | ✅ | | | ✅ | ✅ | | ✅ |

✅ served, blank means Google does not publish the model at that location. The gateway proxies customer traffic that already runs in the customer's own GCP project, so the catalog serves and prices every region Google publishes, including the two APAC regions. Priced regions and served locations are separate facts: the only invariant is that every priced region must be a served location, guarded by `TestGeminiRegionalOverride` / `TestClaudeRegionalOverride`.

## How to Verify

Run from the repo root - it pins its own Go toolchain:

```bash
task test
task lint
```

## Related Resources

- Design: [RFC-0014 Vertex AI provider](https://github.com/redpanda-data/cloudv2/blob/main/apps/aigw/docs/rfcs/0014-vertex-ai-provider.md), [milestone M1](https://github.com/redpanda-data/cloudv2/blob/main/apps/aigw/docs/rfcs/0014-vertex-ai-provider.md#-milestones).
- Jira Subtask: [AI-2016](https://redpandadata.atlassian.net/browse/AI-2016) - ai-sdk-go: add the Vertex provider package with its model catalog and rates.
- Jira Epic: [AI-1943](https://redpandadata.atlassian.net/browse/AI-1943) - LLM Provider: VertexAI.

[AI-2027]: https://redpandadata.atlassian.net/browse/AI-2027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/providers/vertex/models.go
+++ b/providers/vertex/models.go
@@ -72,13 +72,13 @@ const (
 // Attribute keys carried on every Vertex offering. Keys are snake_case
 // and values are strings so the committed snapshot stays stable.
 const (
-	// AttrPublisher is the Vertex publisher segment ("google",
+	// ModelMetadataPublisher is the Vertex publisher segment ("google",
 	// "anthropic").
-	AttrPublisher = "publisher"
-	// AttrVertexModel is the bare wire model ID, i.e. the offering ID
-	// with the vertex. prefix stripped. The offering ID is the pricing
+	ModelMetadataPublisher = "publisher"
+	// ModelMetadataVertexModel is the bare wire model ID, i.e. the offering
+	// ID with the vertex. prefix stripped. The offering ID is the pricing
 	// key; this is what goes in the request path.
-	AttrVertexModel = "vertex_model"
+	ModelMetadataVertexModel = "vertex_model"
 )
 
 // catalogID returns the namespaced offering ID for a bare Vertex model.
@@ -212,8 +212,8 @@ func entries() []catalog.Entry {
 			},
 			Pricing: geminiFlashPricing(),
 			Attributes: map[string]string{
-				AttrPublisher:   publisherGoogle,
-				AttrVertexModel: ModelGemini36Flash,
+				ModelMetadataPublisher:   publisherGoogle,
+				ModelMetadataVertexModel: ModelGemini36Flash,
 			},
 		},
 		{
@@ -236,8 +236,8 @@ func entries() []catalog.Entry {
 			},
 			Pricing: claudeSonnet5Pricing(),
 			Attributes: map[string]string{
-				AttrPublisher:   publisherAnthropic,
-				AttrVertexModel: ModelClaudeSonnet5,
+				ModelMetadataPublisher:   publisherAnthropic,
+				ModelMetadataVertexModel: ModelClaudeSonnet5,
 			},
 		},
 		{
@@ -256,8 +256,8 @@ func entries() []catalog.Entry {
 			},
 			Pricing: claudeHaiku45Pricing(),
 			Attributes: map[string]string{
-				AttrPublisher:   publisherAnthropic,
-				AttrVertexModel: ModelClaudeHaiku45,
+				ModelMetadataPublisher:   publisherAnthropic,
+				ModelMetadataVertexModel: ModelClaudeHaiku45,
 			},
 		},
 	}
--- a/providers/vertex/models.go
+++ b/providers/vertex/models.go
@@ -130,8 +130,6 @@ var geminiCaps = llm.ModelCapabilities{
 	Reasoning:        true,
 }
 
-// geminiModalities: text, image, audio, video, and PDF inputs; text
-// output.
 var geminiModalities = catalog.Modalities{
 	Input: []catalog.Modality{
 		catalog.ModalityText, catalog.ModalityImage, catalog.ModalityAudio,
@@ -157,7 +155,6 @@ var claudeCaps = llm.ModelCapabilities{
 	Reasoning:        true,
 }
 
-// claudeModalities: text, image, and PDF inputs; text output.
 var claudeModalities = catalog.Modalities{
 	Input:  []catalog.Modality{catalog.ModalityText, catalog.ModalityImage, catalog.ModalityDocument},
 	Output: []catalog.Modality{catalog.ModalityText},
@@ -213,18 +210,6 @@ func entries() []catalog.Entry {
 			Life: catalog.Lifecycle{
 				Available: catalog.MustDate("2026-07-21"),
 			},
-			// On Vertex, Gemini 3.6 Flash bills at the introductory rate
-			// ($0.75/$3.75, cache $0.075) through 2026-12-31, then the
-			// standard rate ($1.50/$7.50, cache $0.15) from 2027-01-01.
-			// The catalog tracks the rate in effect, matching the
-			// Gemini-API provider (providers/google/models.go). This is a
-			// real per-token price, not the separate 50% Provisioned
-			// Throughput promotional credit the page also lists; that
-			// credit is post-hoc spend accounting the pricing package puts
-			// out of scope (pricing/doc.go). A non-global endpoint bills
-			// ~10% above global across the us/eu multi-regions, expressed
-			// as Region overrides rather than separate model entries.
-			// Rates read from the page on 2026-09-08.
 			Pricing: geminiFlashPricing(),
 			Attributes: map[string]string{
 				AttrPublisher:   publisherGoogle,
@@ -238,8 +223,8 @@ func entries() []catalog.Entry {
 			Modalities:   claudeModalities,
 			Constraints: llm.ModelConstraints{
 				TemperatureRange: [2]float64{0.0, 1.0},
-				MaxInputTokens:   1000000, // 1M context window
-				MaxOutputTokens:  128000,  // 128K output tokens
+				MaxInputTokens:   1000000,
+				MaxOutputTokens:  128000,
 				SupportedParams:  []string{"temperature", "top_p", "top_k", "max_tokens", "reasoning_effort"},
 			},
 			Reasoning: catalog.ReasoningSupport{
@@ -249,14 +234,6 @@ func entries() []catalog.Entry {
 			Life: catalog.Lifecycle{
 				Available: catalog.MustDate("2026-06-29"),
 			},
-			// $2/$10 is Sonnet 5's global list price, cache reads at the
-			// 0.10x multiplier and writes at 1.25x (5m) / 2x (1h). A
-			// non-global endpoint bills ~10% above global, the same premium
-			// Gemini carries, expressed as Region overrides on the served
-			// multi-regions. Anthropic is the one publisher absent from
-			// Google's Billing Catalog, so Claude rates rest on Google's
-			// Agent Platform pricing page, verified 2026-09-08 (see
-			// claudeSonnet5Pricing).
 			Pricing: claudeSonnet5Pricing(),
 			Attributes: map[string]string{
 				AttrPublisher:   publisherAnthropic,
@@ -270,16 +247,13 @@ func entries() []catalog.Entry {
 			Modalities:   claudeModalities,
 			Constraints: llm.ModelConstraints{
 				TemperatureRange: [2]float64{0.0, 1.0},
-				MaxInputTokens:   200000, // 200K context window
-				MaxOutputTokens:  64000,  // 64K output tokens
+				MaxInputTokens:   200000,
+				MaxOutputTokens:  64000,
 				SupportedParams:  []string{"temperature", "top_p", "top_k", "max_tokens"},
 			},
 			Life: catalog.Lifecycle{
 				Available: catalog.MustDate("2025-10-15"),
 			},
-			// $1/$5 global list price, cache reads 0.10x, writes 1.25x
-			// (5m) / 2x (1h), with the same ~10% non-global premium as
-			// Sonnet on the served named regions (see claudeHaiku45Pricing).
 			Pricing: claudeHaiku45Pricing(),
 			Attributes: map[string]string{
 				AttrPublisher:   publisherAnthropic,
@@ -292,6 +266,12 @@ func entries() []catalog.Entry {
 // geminiFlashPricing builds the Gemini 3.6 Flash rate card: the global
 // rate as the default, and the ~10% higher non-global rate as one Region
 // override per multi-region that carries the premium.
+//
+// The default is Gemini 3.6 Flash's introductory per-token rate, tracked
+// like the Gemini-API provider (providers/google/models.go). It is a real
+// per-token price, not the separate 50% Provisioned Throughput credit the
+// page also lists; that credit is post-hoc spend accounting the pricing
+// package excludes (pricing/doc.go).
 //
 // The priced regions are their own literal, not derived from the
 // availability matrix (locations.go). Pricing and availability are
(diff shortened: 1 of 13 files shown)


providers/vertex/models.go
```diff
@@ -0,0 +1,386 @@
...
+// doc for why the prefix is required rather than optional.
+const catalogKeyPrefix = "vertex."
+
+// Bare Vertex model IDs - exactly the model segment of a Vertex resource
+// path, publishers/{publisher}/models/{model}. The day-one catalog is
+// Gemini + Claude, enumerated in RFC-0014's Model catalog and pricing
+// section (the F5 scope call). The open-weight families on the
+// OpenAI-compatible route are deferred until a customer asks.
+const (
+	ModelGemini36Flash = "gemini-3.6-flash"
```
weeco, line 58: Can we add the newer 3.7 and 3.8 flash models here?

weeco: We should also add the latest Opus and Fable models I guess

providers/vertex/models.go
```diff
@@ -0,0 +1,386 @@
...
+// Rates are USD-per-million-tokens. Every rate here was transcribed from
+// Google's published Vertex pricing; re-reading each one against the live
+// pages is an M1 exit condition, because a catalog PR is the last moment a
+// wrong rate costs nothing.
+//
+// Every rate, Gemini and Claude alike, comes from Google's one pricing
+// page for the platform (Google renamed Vertex AI to the Gemini Enterprise
+// Agent Platform, so older /vertex-ai/ links redirect here):
+//
+//	https://cloud.google.com/gemini-enterprise-agent-platform/generative-ai/pricing
```
weeco, line 196: Good to have it documented, can we add this to the model reconcile skill so that this is always checked as the primary source? Possibly need to teach the skills the new provider

providers/vertex/models.go
```diff
@@ -0,0 +1,386 @@
...
+}
+
+// OfferingForModel returns the Vertex offering for a bare publisher model
+// ID, adding the vertex. catalog-key prefix and looking it up. It is the
+// bridge for callers that hold a bare model name rather than a namespaced
+// offering ID, keeping the prefix an internal catalog detail. A model ID
+// that already carries the prefix is accepted as-is. ok is false for a
+// model the catalog does not offer.
+func OfferingForModel(model string) (catalog.Offering, bool) {
+	return Catalog().Lookup(catalogID(bareModelID(model)))
```
weeco: Probably should use the `Resolve()` instead so that we can match aliases too

providers/vertex/models.go
```diff
@@ -0,0 +1,386 @@
...
+			},
+			// Vertex's own GA date for this model is not published; the zero
+			// value means "available, exact date unknown" (catalog/lifecycle.go).
+			Life:    catalog.Lifecycle{},
+			Pricing: geminiFlashPricing(),
+			Attributes: map[string]string{
+				ModelMetadataPublisher:   publisherGoogle,
+				ModelMetadataVertexModel: ModelGemini36Flash,
+			},
+		},
```
weeco, line 231: I think this is missing the thinking/reasoning efforts: https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/thinking

providers/vertex/models.go
```diff
@@ -0,0 +1,386 @@
...
+			Modalities:   geminiModalities,
+			Constraints: llm.ModelConstraints{
+				TemperatureRange: [2]float64{0.0, 2.0},
+				MaxInputTokens:   1048576, // 1M input tokens
+				MaxOutputTokens:  65536,   // 64K output tokens
+				SupportedParams:  geminiParams,
+			},
+			// Vertex's own GA date for this model is not published; the zero
+			// value means "available, exact date unknown" (catalog/lifecycle.go).
+			Life:    catalog.Lifecycle{},
```
weeco: Release dates are documented here: https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/gemini/3-6-flash

providers/vertex/models.go
```diff
@@ -0,0 +1,386 @@
...
+				MaxOutputTokens:  128000,
+				SupportedParams:  []string{"temperature", "top_p", "top_k", "max_tokens", "reasoning_effort"},
+			},
+			Reasoning: catalog.ReasoningSupport{
+				Efforts:  []llm.ReasoningEffort{reasoningEffortLow, reasoningEffortMedium, reasoningEffortHigh, reasoningEffortXHigh, reasoningEffortMax},
+				Adaptive: true,
+			},
+			// Vertex's own GA date for this model is not published; the zero
+			// value means "available, exact date unknown" (catalog/lifecycle.go).
+			Life:    catalog.Lifecycle{},
```
weeco: Release dates for claude models is also documented: https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/partner-models/claude/sonnet-5

providers/vertex/provider.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// Provider is the Google Vertex AI provider's catalog surface: its name
+// and its validated model catalog. It satisfies catalog.Provider, which
+// is what the spending pipeline and the catalog snapshot consume.
+//
+// It carries no request transport yet. Constructing Vertex requests is a
+// later milestone (RFC-0014 M8); until then the provider contributes its
+// catalog and rates, and the AI Gateway forwards the customer's own
+// native Vertex requests.
```
weeco: Can we shorten all the comments in this PR - more concise and removing irrelevant parts (probably Opus 5 comments here). ANd we can also tighten the test surface a bit IMO. I think the repetitive RFC references are noise

weeco (commented):
I think we can stick to vertex as name

providers/vertex/models.go
MaciejWrzesniewski-Redpanda: I would prefer to keep this PR to the three M1 models if that is ok with you. The three M1 models are not an arbitrary subset, they are the smallest set that exercises both catalog paths end to end, one Gemini model and both Claude reasoning shapes, so the provider mechanics are proven first. Once that flow is proven I will add the remaining Gemini and Claude models, including the newer flash models and the latest Opus and Fable, as originally scoped, under [AI-2108](https://redpandadata.atlassian.net/browse/AI-2108).

providers/vertex/models.go
MaciejWrzesniewski-Redpanda: I agree, and I have added a Vertex sources block to the reconcile-models skill so the Agent Platform pricing and per-model pages are checked as the primary source. I have also taught both skills the new provider, so I added the same Vertex line to the add-model skill sources.

providers/vertex/models.go
MaciejWrzesniewski-Redpanda: I agree, and I have switched OfferingForModel to use Resolve instead of Lookup so it matches aliases too.

providers/vertex/models.go
MaciejWrzesniewski-Redpanda: You are right, I was missing those. I have added the thinking efforts to the Gemini 3.6 Flash entry (minimal, low, medium, high), matching the set the Gemini API provider already uses. I have also added the reasoning efforts to the Claude Haiku 4.5 entry, which was missing them too. Claude Sonnet 5 already carried them.

providers/vertex/models.go
MaciejWrzesniewski-Redpanda: Thank you, I have set the Available date from that page to 2026-07-21. The page does not publish a retirement date, only the earliest date Google could retire the model, which is not a scheduled shutdown. Our Retires field is for an announced shutdown date, so I have left it unset until Google publishes one.

providers/vertex/models.go
MaciejWrzesniewski-Redpanda: Thank you, I have used that page for the Claude dates too, so Available is 2026-06-30 for Sonnet 5 and 2025-10-15 for Haiku 4.5. Both pages give only the earliest date Google could retire the models, not a scheduled shutdown, so I have left Retires unset until a real date is published.

providers/vertex/provider.go
MaciejWrzesniewski-Redpanda: For context, I am using Opus 4.8 with rules to be brief. I also reviewed those comments twice, with second round of a swarm of reviewers cutting out lines out and checking whether valuable information was lost. My point is that it was a much worse issue before, and I worked hard on it, I can swear 😅 

- I agree, and I have shortened the comments across the PR and removed the repetitive RFC references.
- I have also trimmed the Opus 5 mention you flagged in the provider doc.
- Before I go further I would like to check two things with you: which Opus 5 comments did you have in mind beyond that one, and which tests would you like me to tighten. I am happy to cut more once I know where you would like me to focus.

weeco: Overall comments could be more concise still I believe. Give it a shot with Fable 5.1, which seems to be decent for this imo. The comments for the different pricing funcs `func claudeHaiku45Pricing() pricing.Info {` or `func claudeSonnet5Pricing() pricing.Info {` are the examples I had in mind. 

Here are some examples for different places:


```suggestion
// TODO(maciej): add request transport; see RFC-0014 and #<issue>.

// Provider exposes the model catalog for Google's Gemini Enterprise Agent
// Platform (formerly Vertex AI). It implements catalog.Provider but does
// not yet send requests; callers that need a transport should use the
// native Google client directly.
type Provider struct{}
```


Or for the other areas:

```go
// geminiFlashPricing returns the Gemini 3.6 Flash rate card: the global
// rate as default, and the 10% higher non-global rate as one override per
// priced multi-region.
func geminiFlashPricing() pricing.Info {

// claudeSonnet5Pricing returns the Sonnet 5 rate card: the global rate as
// default, and Google's flat 10% non-global premium as one override per
// served non-global region.
func claudeSonnet5Pricing() pricing.Info {

// claudeHaiku45Pricing returns the Haiku 4.5 rate card, same shape as
// [claudeSonnet5Pricing].
func claudeHaiku45Pricing() pricing.Info {
```

```go
// NewProvider returns a provider for Google's Gemini Enterprise Agent
// Platform (formerly Vertex AI).
func NewProvider(_ context.Context) (*Provider, error) {
      return &Provider{}, nil
}
```

providers/vertex/models.go
weeco: Okay fine for me, was just surprised that we used the latest Haiku/Sonnet models, but not the latest flash models

providers/vertex/provider.go
MaciejWrzesniewski-Redpanda: great, all applied, please re-check 🚀

weeco (approved):

PR #222, opened 2026-09-09
fix: low-cardinality error.type on spans, no doubled "API call failed" prefix

mmaslankaprv:
## What a failed model call looked like downstream

An agent whose LLM call was refused by the gateway (a guardrail block, HTTP 403) produced this turn in Redpanda's transcripts UI:

```
Turn failed · *fmt.wrapErrors
agent: model generation failed: API call failed: API call failed: [guardrail_intervened] Blocked by the demo guardrail: this topic is off limits.
```

Two defects in that line come from this SDK.

## 1. `error.type` = Go type of the error

`plugins/otel` `setSpanError` set `error.type` to `fmt.Sprintf("%T", err)`. Any wrapped error is one of the stdlib wrapper types, so the attribute was `*fmt.wrapErrors` / `*fmt.wrapError` / `*errors.errorString` — no information, and shown verbatim as the error code in transcripts.

Now `spanErrorType`:
- a `*llm.ProviderError` anywhere in the chain contributes its provider **code** (`rate_limit_exceeded`, `guardrail_intervened`) — the low-cardinality class the semantic conventions ask for;
- other concrete types keep their type name as before;
- the stdlib wrappers (`*fmt.wrapError`, `*fmt.wrapErrors`, `*errors.errorString`, `*errors.joinError`) record **no** `error.type`.

`tool_error` for tool-payload errors is untouched.

## 2. "API call failed: API call failed:"

Every provider wrapped the classified error with `ErrAPICall` a second time ("so both `errors.Is` checks work"). `classifyStatusCode` maps 401/403 to `ErrAPICall` itself, so the `*ProviderError` already rendered `API call failed: [code] message` and the outer wrap doubled the prefix.

`llm.WrapAPICall(err)` adds the category only when `errors.Is(err, ErrAPICall)` does not already hold. All eleven call sites (`anthropic`, `bedrock`, `google`, `openai`, `openaicompat`; `Generate` + `GenerateEvents`) use it. `errors.Is(err, ErrAPICall)` and `errors.Is(err, <specific>)` both still hold — pinned by the test.

## Tests

`llm/errors_wrap_test.go`, `plugins/otel/utils_errortype_test.go`. `task lint:new`, `task license:check`, unit tests for `llm`, `providers`, `plugins`, `agent`, `adapter` green.

Consumer-side follow-up (cloudv2): the transcripts assembler will prefer the step's error when the run's error merely wraps it, dropping the `agent: model generation failed:` prefix; then a go.mod bump to this change.

### DIFF
--- a/llm/errors.go
+++ b/llm/errors.go
@@ -106,6 +106,20 @@ type ProviderError struct {
 	Retryable bool
 }
 
+// WrapAPICall marks err as an API-call failure so errors.Is(err, ErrAPICall)
+// holds for callers that only check the category, while keeping the specific
+// classification (rate limit, context overflow, ...) reachable with errors.Is
+// too. A classified error whose category already is ErrAPICall — a
+// *ProviderError for a 401/403 — is returned as is: wrapping it again would
+// repeat the "API call failed" prefix in the message.
+func WrapAPICall(err error) error {
+	if err == nil || errors.Is(err, ErrAPICall) {
+		return err
+	}
+
+	return fmt.Errorf("%w: %w", ErrAPICall, err)
+}
+
 // IsRetryable checks if an error represents a transient condition that may
 // succeed on retry. It works through error wrapping chains.
 //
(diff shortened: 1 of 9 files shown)


birdayz (approved):

alenkacz (approved):
LGTM

PR #223, opened 2026-09-14
feat(pricing)!: Key the catalog by provider and model (AI-2118)

MaciejWrzesniewski-Redpanda:
## Summary

The pricing catalog keyed every rate card by model ID alone. But a model ID is not unique across providers: for instance, `claude-sonnet-5` is sold by both Anthropic and Vertex, at different rates. One key per ID means one rate card per ID, so the second provider to register a shared ID cannot get its own rates in.

This PR keys by the pair `{provider, model}` instead. A shared ID now holds one rate card per provider.

## Merge Order Across the Two Repos

The order is fixed: cloudv2 cannot compile against the new `Lookup` and `Calculate` signatures until they exist on an ai-sdk-go commit.

1. **Review and approve this PR, then merge it to `main`.**
2. **[cloudv2 #29947](https://github.com/redpanda-data/cloudv2/pull/29947) re-pins its `go.mod` onto the merged commit.**

[AI-2020](https://redpandadata.atlassian.net/browse/AI-2020) sets the outer deadline: dropping the `vertex.` prefix is a plain revert only while no stored row carries a prefixed Vertex ID, and that work writes the first such rows. Neither half of this change touches a schema, a stored column, or a rollup row, so there is no migration and no backfill.

## Review Load

| Bucket | Files # | Code Lines | Comment Lines |
|:---|:---:|:---:|:---:|
| Implementation | 39 | 229 | 164 |
| Tests | 21 | 377 | 94 |
| Docs (`MIGRATION.md`) | 1 | 77 | N/A |
| Generated (`catalog/snapshot.json`) | 1 | +367 / -379 | N/A |

## Scope

- The catalog version hash changes value for every model: it moves to schema `v2` and folds in the provider and the set of registered providers. Versions stored before this release stay valid for historical rows and will not recur.
- The Google catalog's registration key changes value from `google` to `gcp.gemini`, matching the name that provider reports.
- Vertex offering IDs lose their `vertex.` namespace and become the bare publisher IDs. The internal prefix-stripping in the Vertex location helpers goes with it, so a caller still passing a prefixed ID compiles and gets a miss rather than an error.
- `catalog/snapshot.json` is regenerated, so the read format non-Go consumers depend on changes under them.
- Each change above breaks a consumer without breaking its build: they change values, not signatures. `MIGRATION.md` lists them with the consequence of missing each.
- Two of the breaks are not like those four: they are compile errors, not changed values. `vertex.ModelMetadataVertexModel` and the `vertex_model` attribute it keyed are removed - the attribute held the offering ID with the `vertex.` prefix stripped, which the bare IDs above already are, so it duplicated `Offering.ID` on every Vertex entry, and non-Go consumers lose the entry from each Vertex offering's `attributes` list. `vertex.OfferingForModel` is removed with it - stripping that prefix was its only job, so it had become a verbatim alias of `vertex.Catalog().Resolve`, which callers should use instead.
- The worked examples in `MIGRATION.md`, `pricing/doc.go`, and `catalog/catalog.go` are rewritten to the new call shapes.
- Every provider package's model and provider getter is retyped, and the protobuf session record, the OTel attributes, the snapshot DTO, `agent.Info` and `openaicompat`'s user-supplied name each convert at their own boundary with a comment saying why.

### The three IDs two providers already share

`catalog/snapshot.json` already carries three model IDs under two providers each, and the two rate cards differ every time: `claude-sonnet-5` and `claude-haiku-4-5` under both `anthropic` and `gcp.vertex`, and `gemini-3.6-flash` under both `gcp.gemini` and `gcp.vertex`. Each direct provider charges one flat rate and Vertex charges a regional premium on top, so the pairs cannot collapse into one entry. The rates are in that file.

The old fix was to rename the model rather than key it properly: Vertex's offering ID became `vertex.claude-sonnet-5`, which is not the ID the provider reports. Helpers then had to strip the prefix back off, and the renamed ID leaked into `catalog/snapshot.json`, the file non-Go consumers read. All three IDs above now exist once per provider, under the real ID, each with its own rate card.

### The `{provider, model}` key and the two-error split

The catalog map went from `map[string]Info` to `map[modelKey]Info`, where `modelKey` is `{provider ProviderKey, model string}`, and the builder records each registered provider so a lookup can tell "provider never registered" from "model not found under a known provider":

| Call | Provider registered? | Model found? | Result |
|---|---|---|---|
| `Lookup("gcp.vertex", "claude-sonnet-5")` | yes | yes | `Info`, nil error |
| `Lookup("openai", "made-up")` | yes | no | `ErrUnknownModel` |
| `Lookup("typo-provider", ...)` | no | - | `ErrUnknownProvider` |

### One `ProviderName` constant removes the drift

The Google model catalog was built under `"google"` while `Provider.Name()` returned `"gcp.gemini"`. Once the catalog keys by provider, `WithSource(google.Catalog())` would register under `gcp.gemini` while a hand-typed `"google"` lookup missed and priced at zero. Every provider package that registers a catalog key now exports one `const ProviderName` - `anthropic`, `bedrock`, `google`, `meta`, `openai`, `vertex` - used by both the catalog and `Name()`, and `cmd/catalog-snapshot` asserts the two agree and that the keys are pairwise distinct. `openaicompat` is the exception and needs no constant: its `Catalog()` returns `nil` by design, because the model names come from whatever OpenAI-compatible endpoint it points at, so it registers no key and has no second spelling to drift from.

### One provider type, and why `ProviderKey` is an alias

The constant removes the drift between one provider's two spellings; the type removes the casts between the two catalogs. `llm.ProviderID` is the single named type, and `pricing.ProviderKey` is an alias of it rather than a second type, which is what lets a model catalog's `Provider()` feed a pricing option with no conversion. Both `pricing.ProviderKey(...)` conversions outside tests are gone.

A distinct `ProviderKey` was built and measured first. It costs three casts - `pricing/builder.go:73`, `catalog/catalog.go:225`, and one in cloudv2 that could not be deleted - and a cast is a crossing the compiler does not check, so it puts the original defect back in a smaller form.

### Public Interface Delta

- `github.com/redpanda-data/ai-sdk-go/llm`
  - `api.go`
    - [CHANGED interface method `ModelInfo.Provider` - `func() string` -> `func() ProviderID` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/llm/api.go#L61)
    - [ADDED type `ProviderID`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/llm/api.go#L30)
  - `part.go`
    - [CHANGED struct field `ToolSearchPart.Provider` - `string` -> `ProviderID` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/llm/part.go#L89)
- `github.com/redpanda-data/ai-sdk-go/pricing`
  - `catalog.go`
    - [CHANGED method `Lookup` - `func(string) (Info, bool)` -> `func(ProviderKey, string) (Info, error)` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/catalog.go#L74)
    - [CHANGED method `Calculate` - gained a leading `ProviderKey` parameter **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/catalog.go#L114)
    - [ADDED type alias `ProviderKey` = `llm.ProviderID`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/catalog.go#L30)
    - [ADDED var `ErrUnknownProvider`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/catalog.go#L55)
  - `builder.go`
    - [CHANGED func `WithProvider` - `func(string, map[string]Info) Option` -> `func(ProviderKey, map[string]Info) Option` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/builder.go#L82)
    - [CHANGED func `WithOverride` - `func(string, Info) Option` -> `func(ProviderKey, string, Info) Option` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/builder.go#L119)
    - [ADDED interface `Source`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/builder.go#L42)
    - [ADDED func `WithSource`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/builder.go#L55)
- `github.com/redpanda-data/ai-sdk-go/catalog`
  - `catalog.go`
    - [CHANGED func `New` - first parameter `string` -> `llm.ProviderID` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/catalog/catalog.go#L89)
    - [CHANGED func `MustNew` - first parameter `string` -> `llm.ProviderID` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/catalog/catalog.go#L241)
    - [CHANGED method `(*Catalog).Provider` - `func() string` -> `func() llm.ProviderID` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/catalog/catalog.go#L344)
  - `offering.go`
    - [CHANGED method `Offering.Provider` - `func() string` -> `func() llm.ProviderID` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/catalog/offering.go#L40)
  - `provider.go`
    - [CHANGED interface method `Provider.Name` - `func() string` -> `func() llm.ProviderID` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/catalog/provider.go#L31)
- `github.com/redpanda-data/ai-sdk-go/catalog/snapshot`
  - `snapshot.go`
    - [CHANGED const `SchemaVersion` - `1` -> `2` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/catalog/snapshot/snapshot.go#L51)
- `github.com/redpanda-data/ai-sdk-go/llm/fakellm`
  - `fake.go`
    - [CHANGED method `(*FakeModel).Provider` - `func() string` -> `func() llm.ProviderID` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/llm/fakellm/fake.go#L238)
- `github.com/redpanda-data/ai-sdk-go/providers/{anthropic,bedrock,google,meta,openai,openaicompat,vertex}`
  - `provider.go` - `(*Provider).Name` retyped `func() string` -> `func() llm.ProviderID`, **breaking** in every package
    - [CHANGED method `(*Provider).Name` - `anthropic`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/anthropic/provider.go#L58)
    - [CHANGED method `(*Provider).Name` - `bedrock`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/bedrock/provider.go#L140)
    - [CHANGED method `(*Provider).Name` - `google`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/google/provider.go#L43)
    - [CHANGED method `(*Provider).Name` - `meta`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/meta/provider.go#L46)
    - [CHANGED method `(*Provider).Name` - `openai`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/openai/provider.go#L42)
    - [CHANGED method `(*Provider).Name` - `openaicompat`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/openaicompat/provider.go#L43)
    - [CHANGED method `(*Provider).Name` - `vertex`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/vertex/provider.go#L42)
  - `model.go` - `(*Model).Provider` retyped `func() string` -> `func() llm.ProviderID`, **breaking** in every package
    - [CHANGED method `(*Model).Provider` - `anthropic`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/anthropic/model.go#L51)
    - [CHANGED method `(*Model).Provider` - `bedrock`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/bedrock/model.go#L52)
    - [CHANGED method `(*Model).Provider` - `google`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/google/model.go#L53)
    - [CHANGED method `(*Model).Provider` - `openai`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/openai/model.go#L50)
    - [CHANGED method `(*Model).Provider` - `openaicompat`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/openaicompat/model.go#L48)
  - `models.go`
    - [ADDED const `ProviderName` = `anthropic`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/anthropic/models.go#L82)
    - [ADDED const `ProviderName` = `aws.bedrock`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/bedrock/models.go#L560)
    - [ADDED const `ProviderName` = `gcp.gemini`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/google/models.go#L47)
    - [ADDED const `ProviderName` = `meta`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/meta/models.go#L35)
    - [ADDED const `ProviderName` = `openai`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/openai/models.go#L26)
    - [ADDED const `ProviderName` = `gcp.vertex`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/vertex/models.go#L41)
- `github.com/redpanda-data/ai-sdk-go/providers/vertex`
  - `models.go`
    - [DELETED const `OfferingGemini36Flash` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/d6d5e65e01924e307ed36c254a97222a9ad5a593/providers/vertex/models.go#L69)
    - [DELETED const `OfferingClaudeSonnet5` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/d6d5e65e01924e307ed36c254a97222a9ad5a593/providers/vertex/models.go#L70)
    - [DELETED const `OfferingClaudeHaiku45` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/d6d5e65e01924e307ed36c254a97222a9ad5a593/providers/vertex/models.go#L71)
    - [DELETED const `ModelMetadataVertexModel` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/d6d5e65e01924e307ed36c254a97222a9ad5a593/providers/vertex/models.go#L90) - with it, the `vertex_model` entry in every Vertex offering's `attributes`
    - [DELETED func `OfferingForModel` **breaking**](https://github.com/redpanda-data/ai-sdk-go/blob/d6d5e65e01924e307ed36c254a97222a9ad5a593/providers/vertex/models.go#L109) - it had become a verbatim alias of `Catalog().Resolve`; call that instead

### Acceptance Criteria

- **AC 1** - two sources whose catalogs share a model ID each price under their own rate card, and the duplicate error fires only for the same provider and model.
  - [`TestBuilder_SharedModelIDAcrossProviders`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/pricing_test.go#L317)
  - [`TestWithProvider_SameProviderInTwoPieces`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/pricing_test.go#L342) - the duplicate error does not fire when one provider is registered in two pieces
- **AC 2** - `Lookup` and `Calculate` require a provider as well as a model, so a provider-less call does not compile, and a source with an empty provider name is rejected when the catalog is built.
  - [`TestWithSource_EmptyProviderFailsBuild`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/pricing_test.go#L614)
  - [`TestWithProvider_EmptyKeyFailsBuild`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/pricing_test.go#L667)
- **AC 3** - Vertex offering IDs in `catalog/snapshot.json` carry no `vertex.` prefix, and `claude-sonnet-5` returns a different rate card under `gcp.vertex` than under `anthropic`.
  - [`TestNoNamespacedPricingKey`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/providers/vertex/models_test.go#L71)
  - [`TestVertexAndAnthropicPriceSonnet5Differently`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/cmd/catalog-snapshot/lifecycle_test.go#L77) - the work item names this one `TestVertexPricingDoesNotCollide`
  - [`TestCommittedSnapshotIsFresh`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/cmd/catalog-snapshot/snapshot_test.go#L33) - the committed `catalog/snapshot.json` matches what the generator emits, prefix included
- **AC 4** - every provider package's catalog reports the provider name that provider reports, and the Google catalog reports `gcp.gemini`.
  - [`TestCatalogProviderMatchesProviderName`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/cmd/catalog-snapshot/provider_name_test.go#L41)
  - [`TestProviderKeysArePairwiseDistinct`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/cmd/catalog-snapshot/provider_name_test.go#L75) - no two providers claim the same key
- **AC 5** - `spending_pricing_unknown_total` carries `provider` and `model` labels, and the metrics inventory says so.
  - closed by: [cloudv2 #29947](https://github.com/redpanda-data/cloudv2/pull/29947)
- **AC 6** - a gateway consumer registering two providers with one shared model ID prices each event under its own provider.
  - closed by: [cloudv2 #29947](https://github.com/redpanda-data/cloudv2/pull/29947)
- **AC 7** - an unregistered provider and an unknown model under a known provider report different errors, and a hit returns no error.
  - [`TestLookup_UnknownProvider`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/pricing_test.go#L273) - also asserts the nil error on a hit
  - [`TestCalculate_UnknownModelReturnsError`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/pricing_test.go#L293)

`task test` runs every test above, the snapshot freshness check included, so CI covers this section end to end.

### Additional Criteria

Authored for this change, not carried by [AI-2118](https://redpandadata.atlassian.net/browse/AI-2118).

- **A** - a typed-nil provider catalog is rejected when the catalog is built, rather than panicking there or at the first lookup.
  - [`TestTypedNilSatisfiesPricingSource`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/catalog/catalog_test.go#L660) - asserted from the consuming package, where a typed nil actually arises
- **B** - a lookup against a nil catalog reports an unknown provider instead of panicking.
  - [`TestNilCatalog_ReportsUnknownProvider`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/pricing_test.go#L761)
- **C** - an override naming an empty provider key is reported as that, not as an override of an unknown model.
  - [`TestWithOverride_EmptyProviderFailsBuild`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/pricing_test.go#L678)
- **D** - a provider registered with no models still changes the catalog version, because an empty-map provider is known and changes how a miss is classified.
  - [`TestCatalogVersion_FoldsInKnownProviderSet`](https://github.com/redpanda-data/ai-sdk-go/blob/bc8a11809eab1e259750eef542793bff4a0712b0/pricing/pricing_test.go#L649)

## Related Resources

- [AI-2118](https://redpandadata.atlassian.net/browse/AI-2118) - the story (Agent Experience; two-repo change, ai-sdk-go lands first)
- [cloudv2 #29947](https://github.com/redpanda-data/cloudv2/pull/29947) - the gateway half, reviewed at the same time and merged after this one
- [ai-sdk-go #221](https://github.com/redpanda-data/ai-sdk-go/pull/221) - introduced the `vertex.` prefix and the exported `Offering*` consts this PR removes
- [AI-2020](https://redpandadata.atlassian.net/browse/AI-2020), [AI-2108](https://redpandadata.atlassian.net/browse/AI-2108) - the Vertex follow-ups

[AI-2118]: https://redpandadata.atlassian.net/browse/AI-2118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AI-2020]: https://redpandadata.atlassian.net/browse/AI-2020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


catalog/catalog.go
```diff
@@ -220,7 +220,8 @@ func New(provider string, entries []Entry, opts ...Option) (*Catalog, error) {
 
 	// Pricing is validated by the pricing builder, which enforces rate
 	// sanity and override consistency; its errors carry the model ID.
-	if _, err := pricing.NewCatalog(pricing.WithProvider(provider, pricingMap(c.offerings))); err != nil {
+	// WithProvider, not WithSource: c is still mid-construction.
+	if _, err := pricing.NewCatalog(pricing.WithProvider(pricing.ProviderKey(provider), pricingMap(c.offerings))); err != nil {
```
weeco: WDYT about introducing a `llm.ProviderID` custom type here and use that across both catalogs (model catalog & pricing). I think the `llm` package would work for the existing dependency directions (no cycles). 

In the model catalog we use an `OfferingID` for lookups. I think a custom type would make it more robust since you can't have an empty result for `pricing.ProviderKey()` lookups. Plus we should make it consistent across both catalogs. `catalog.New()` and the functions in the `pricing` pkg would both use that new type

pricing/catalog.go
```diff
@@ -22,33 +22,50 @@ import (
...
 
-	info, ok := c.models[modelID]
-	if !ok {
-		return Info{}, false
+// ErrUnknownProvider is returned from Lookup and Calculate when the
+// catalog carries no rates for the requested provider at all. Distinct
+// from ErrUnknownModel: it is almost always a mapping bug — a call site
+// passing a key the catalog was never registered under — and it prices
+// every request there as a silent zero, so alert on it.
+var ErrUnknownProvider = errors.New("pricing: unknown provider")
```
weeco: NIT: Lets group the sentinel errors please

pricing/catalog.go
```diff
@@ -22,33 +22,50 @@ import (
...
+}
 
-	info, ok := c.models[modelID]
-	if !ok {
-		return Info{}, false
+// ErrUnknownProvider is returned from Lookup and Calculate when the
+// catalog carries no rates for the requested provider at all. Distinct
+// from ErrUnknownModel: it is almost always a mapping bug — a call site
+// passing a key the catalog was never registered under — and it prices
+// every request there as a silent zero, so alert on it.
```
weeco: Comment should be more concise - obviously this is distinct from ErrUnknownModel I guess? Seems like its kinda documenting the migration parts of it

providers/google/models.go
```diff
@@ -42,8 +42,12 @@ const (
 	ModelGemini25FlashLite   = "gemini-2.5-flash-lite"
 )
 
+// ProviderName is the catalog key Provider.Name() returns; see
+// pricing.ProviderKey. This pair drifted ("google" vs "gcp.gemini") before AI-2118.
+const ProviderName = "gcp.gemini"
```
weeco: NIT: Shouldn't have private jira ticket references or unnecessary code evolution comments

providers/vertex/models.go
```diff
@@ -85,29 +65,15 @@ const (
 	// ModelMetadataPublisher is the Vertex publisher segment ("google",
 	// "anthropic").
 	ModelMetadataPublisher = "publisher"
-	// ModelMetadataVertexModel is the bare wire model ID (the offering ID
-	// without the vertex. prefix), which goes in the request path.
+	// ModelMetadataVertexModel is the bare wire model ID for the request
+	// path. It equals the offering ID today; AI-2108 drops this attribute.
```
weeco: NIT: No jira ticket refs please

catalog/snapshot/snapshot.go
```diff
@@ -40,7 +40,15 @@ import (
 // rename, removal, or semantic change of an existing field — additions
 // are not version bumps. Tolerant-reader contract: consumers MUST ignore
 // unknown fields and MUST NOT require optional ones.
-const SchemaVersion = 1
+//
+// v2 (AI-2118): the value domain of "id"/"model" changed without the
```
weeco: NIT: No jira ticket refs

MIGRATION.md
```diff
@@ -83,10 +104,55 @@ shape.
...
+  not `vertex.claude-sonnet-5`), and `vertex.Offering*` is now
+  `vertex.Model*`. `OfferingForModel`, `LocationsForModel` and
+  `IsModelAvailableAtLocation` no longer strip a `vertex.` prefix: they
+  take a `string`, so a caller still passing the prefixed form compiles
+  clean and gets `ok == false` / `nil` / `false` — a silently missing
+  model, not a compile error. `catalog/snapshot.json` is the read format
+  for non-Go consumers, so these IDs change under them. Consumers that
+  persisted a prefixed ID must rewrite it before bumping; after the
+  first stored row this becomes a data migration, not a revert. The
+  counterpart change in the AI Gateway is tracked as AI-2020, which
```
weeco: NIT: No jira ticket refs, this is a public repo and code should be self-contained without external references especially when they are private.

pricing/catalog.go
```diff
@@ -22,33 +22,50 @@ import (
...
-// builder errors out if two providers try to register the exact same
-// ID so runtime lookups are always unambiguous.
-type Catalog struct {
-	models  map[string]Info
-	version string
+// ProviderKey is the provider half of a catalog key: the name a provider
+// returns from Name() ("openai", "gcp.gemini", "aws.bedrock", ...). Take
+// it from the provider's ProviderName const rather than hand-typing it;
+// see ErrUnknownProvider for what a wrong key costs.
+type ProviderKey string
```
weeco: Just noticed we have this type here, but aren't using it in the constructor. See my other comment, I think we should have a proper provider type in the llm package instead so that we can share it across packages without cyclic dependencies. 

We also have openai.ProviderName , we have a provider type with a Name() method and now this Providerkey concept. I think we should just have one type, but maybe I am missing something. It's a human review :D

catalog/catalog.go
MaciejWrzesniewski-Redpanda: Thanks Martin, agreed. Done in 3ab86fc.

`llm.ProviderID` is the one named type now, and the model catalog and the pricing catalog both use it. `catalog.New`, `MustNew`, `Catalog.Provider()`, `Offering.Provider()` and the `Provider` interface's `Name()` all take or return it, as do the 7 `Provider.Name()` and 8 `Model.Provider()` implementations.

Worth knowing: the type only helps where nobody writes a cast. `ProviderID(someString)` compiles whatever is in the string. So the gain came from deleting casts, not from adding the type - both `pricing.ProviderKey(` conversions outside tests are gone, this one and `pricing/builder.go:73` in `WithSource`.

What remains converts at a boundary that has to stay primitive - the protobuf session record, the OTel attributes, the snapshot DTO, `agent.Info`, and `openaicompat`'s user-supplied name. Each has a comment saying why.

Two things the type does not buy, for the record: it does not stop an empty value, and it does not catch a wrong-but-plausible key. `llm.ProviderID("")` and `llm.ProviderID("gemini")` both compile and both price at zero.

pricing/catalog.go
MaciejWrzesniewski-Redpanda: Same commit, 3ab86fc. `pricing.ProviderKey` is now `type ProviderKey = llm.ProviderID`.

I tried the distinct type first. It costs three conversions: `pricing/builder.go:73` and `catalog/catalog.go:225`, both of which this commit deletes, plus one on the cloudv2 side that would have to stay. Nothing else breaks, because the `ProviderName` values are untyped constants and assign to either type for free.

Three casts is nothing in lines, but a cast is a crossing the compiler does not check, so a distinct type would put the same defect back in a smaller form. An alias has no crossings at all.

The name stays `ProviderKey` because `pricing` calls it a "provider key" throughout its docs and errors.

pricing/catalog.go
MaciejWrzesniewski-Redpanda: Done in 40a32ea. Both sentinels are in one `var` block above `Lookup` now.

pricing/catalog.go
MaciejWrzesniewski-Redpanda: Done in 40a32ea. Dropped the `Distinct from ErrUnknownModel` clause and the aside after it - a wrong key and a key that was never registered are the same thing, so it said it twice.

I kept the silent-zero line. That is the one thing a caller cannot read off the signature, and it is why this error wants an alert rather than a log.

providers/google/models.go
MaciejWrzesniewski-Redpanda: Done in 0664d61. Dropped the whole sentence, not just the key - `This pair drifted ... before AI-2118` was describing the change rather than the code, and the git log already has that.

providers/vertex/models.go
MaciejWrzesniewski-Redpanda: Done in 0664d61. Same treatment as the Google one: the clause was a note about a future change, so it went with the key.

catalog/snapshot/snapshot.go
MaciejWrzesniewski-Redpanda: Done in 0664d61. `v2 (AI-2118):` is `v2:` now. The rest of that comment says what changed between v1 and v2 without naming a ticket.

MIGRATION.md
MaciejWrzesniewski-Redpanda: Agreed, and thanks for spelling out the reason - a public repo's comments have to stand alone. Done in 0664d61. The gateway counterpart sentence is gone.

`git grep -E 'AI-[0-9]+'` is clean across the repo now except three lines in `runner/runner_test.go`, which are on main already. Not touching them here, but they are the same objection if you want them filed.

MaciejWrzesniewski-Redpanda:
@claude review

MaciejWrzesniewski-Redpanda:
Went through all six sections. Five changes landed, two rejected. Commits `f76a05f`, `e602e12`, `e1b9c29`, `afae09f`, `6f4bb9a`.

**1. `meta` has no `ProviderName` const** - right, and undercounted. There were three bare `"meta"` literals, not two: `providers/meta/models.go:35`, `providers/meta/provider.go:46`, and `providers/meta/provider.go:91` in `metaModel.Provider()`. All three use the const now. `afae09f`.

**1b. The same section also asks for one on `openaicompat`'s default** - no. `Catalog()` returns `nil` there by design, so openaicompat registers no catalog key at all, and its `"openaicompat"` is a fallback for a name the user configures. The drift this PR fixes is one provider having a catalog key and a `Name()` that disagree. With no catalog key there is no second spelling, so a const would guard nothing.

**2. `WithOverride` skips the empty-provider check its siblings have** - right. `f76a05f`. The build did already fail, but as `override for unknown model "m" under provider ""`, which points at the model when the mistake is the key. Deleting the new guard makes the test fail on the message rather than on the build, so it pins the diagnostic and not just the failure.

**3. The `known_provider=` fold is unasserted** - right, and this is the one I checked hardest. `TestCatalogVersion_FoldsInProvider` uses two providers that each carry a model, so the per-model `provider=%q` write already separates them, and the fold is not what makes that test pass. `e602e12` adds a case whose two catalogs differ only in a provider registered with an empty map. Delete the fold and it fails with `Should not be: "75f2baf265a44af6"` - they hash identically, which is exactly the claim.

**4. `vertex_model` is a byte-exact duplicate of `id`** - right, and I removed it rather than documenting it. `6f4bb9a`. Worth saying why the field was not always dead weight: on `main` it carried real information, the offering ID with the `vertex.` prefix stripped, and `main`'s own test asserted `strings.TrimPrefix(o.ID, "vertex.")` against it. This PR drops the prefix, so that assert degraded into comparing a field with its own neighbour. Nothing outside tests ever read it - there is no non-test read of `Offering.Attributes` anywhere in `providers/vertex/` - so the doc comment claiming it fed the request path described a use that did not exist. `MIGRATION.md` carries the break, and Go consumers get a compile error rather than a silent change.

**5. `Source`'s comment says `ProviderKey` while its signature says `llm.ProviderID`** - right. `e1b9c29`.

**6. The residual silent-miss surface is narrower than `MIGRATION.md` implies** - no, and the section concedes as much itself. The bullet opens with the ID change and `vertex.Offering*` becoming `vertex.Model*`, which is the compile error, and only then describes the `ok == false` path for a caller still passing a prefixed string. That is already the ordering it asks for.

One caveat on the review itself: it says `go build` and `task test` were denied in its sandbox, so it is static analysis only. I ran `task lint:new` (0 issues), `task catalog:check` and the touched packages on each commit locally.

The description is updated for all of this and repinned at `6f4bb9a`.

weeco (approved):

PR #224, opened 2026-09-14
feat(otel): carry Runner.Run's userID and stamp it as user.id

alenkacz:
## Why

`Runner.Run` has always taken a user id and thrown it away — the parameter was literally:

```go
_ string, // UserID will be used in the future.
```

So an application that knew which human it was running for had no way to get that into its telemetry. Downstream, a transcript built from these spans cannot say who ran a turn.

## What

The invocation gains a namespace for **caller-asserted attributes**: what the application knows about who it is running for. The otel plugin stamps every one of them on every span it opens, and in-process sub-agents inherit them.

- `agent.AttrUserID` (`"user.id"`) is the well-known key. `Runner.Run` writes its `userID` argument there.
- `inv.SetAttribute` / `Attribute` / `Attributes` to write and read; `agent.WithAttributes` to seed at construction.
- `Runner.Run` takes `RunOption`s — `runner.WithAttribute("user.tier", "premium")` — for anything else the caller knows.
- A sub-agent spawned through `agenttool` inherits the whole set, so its spans carry the same attribution as its parent's.

Adding a variadic parameter is source-compatible: every existing `Run` call site, all three examples included, compiles unchanged.

## Four decisions worth reviewing

**Why a namespace separate from metadata.** This is the substance of the review on this PR, and it went through three shapes: a typed `userID` field, then the existing metadata map under a well-known key, then this. The metadata map turns out to be unable to carry it, because the sub-agent case demands propagation and that map cannot be propagated: it is per-invocation plumbing and holds live objects — the otel plugin parks the invocation's own span in it. Copy it to a child and the child would close the parent's span. Attributes are therefore their own namespace, **string-valued by construction**, which is precisely what makes them always safe both to copy into a sub-agent and to stamp wholesale as span attributes. No allowlist, no type switch, no reviewer vigilance required.

**Why sub-agents inherit at all.** Before this, attribution stopped at the root agent: `agenttool` mints a fresh invocation, so every sub-agent span carried nothing. That is the same failure this PR set out to fix — a consumer reading a tool span cannot name its human — one level down and for a whole subtree.

**Why context is the transport.** `Tool.Execute(ctx, args)` is the entire boundary; a tool never sees the calling invocation. So `ContextWithAttributes` / `AttributesFromContext` mirror the existing `ContextWithConversationID` pair, which solved exactly this problem for conversation grouping, down to the "empty inherits" semantics. `llmagent` publishes them once per tool batch on the context it dispatches tool calls with, beside the conversation id; `agenttool` seeds the child from it. Copies in both directions, so neither side can mutate the other's map.

**Why `RunOption`s rather than interceptor-only.** An application could write attributes from its own interceptor, but only options are applied *before* the invocation starts — so only options are present on the root span and visible to sampling. An interceptor necessarily runs after that span has opened, and whether its writes land on the root span at all depends on its order relative to the otel interceptor.

**Why all three span kinds.** Consumers group spans one at a time, so a value present only on the root span is invisible to a consumer reading a tool span. This mirrors why `gen_ai.conversation.id` is already stamped on all three.

## Extra attributes need no bag of their own

The "attributes only the application knows" case — premium vs. normal, tenant, plan — is served by the same path the user id takes, plus the pre-existing `pluginotel.WithAttributeInjector` for anything that has to be computed per span. Nothing here is a second mechanism for that.

## Semantics

The values are **opaque and unverified by design**: an application asserts whatever it knows — an email, an internal id. This SDK never validates them, and nothing here reads them for authorization. Observability-only, and consumers should treat them the same way. Empty keys and values are rejected at every entry point, so an application that asserts nothing leaves no attribute rather than a blank one on every span.

`user.id` is not a `gen_ai.*` key — the GenAI semantic conventions define no user attribute — but it is where the ecosystem converged: OpenInference stamps it, Langfuse's SDK writes it, and ADK imports the same constant. Because the otel plugin now stamps whatever the caller asserted rather than looking for one key, it no longer names `user.id` anywhere: the constant lives in `agent`, and `attrUserID`, the `userID()` helper and `invocationUserID()` are all gone.

## Drive-by

`SetMetadata` / `GetMetadata` docs now describe metadata as per-invocation plumbing that does not reach telemetry and is not inherited, and point at `SetAttribute` for anything describing the caller. `SetMetadata`'s example was itself an ad-hoc `"user_id"` string.

Two orphaned doc comments fixed: a test had been inserted between `collectEvents`' doc comment and `collectEvents` itself, and the same happened to `conversationCompacted` while adding a helper beside it.

## Merged main

`main` picked up #220 (lazy tool loading via `tool_search`), which restructured `executeTools`: tool calls now run on a worker pool through `a.executeTool`, so the per-call `toolCtx` this branch had hooked into no longer exists, and `ContextWithConversationID` moved with it — hoisted out of the loop to apply once to the function's `ctx`. That was the only conflict, in the only file where these two changes overlap.

Resolved by taking main's structure and re-applying the attribute publication at the hoisted site, beside the conversation id it belongs with — which is also where it wants to be: set once per tool batch rather than rebuilt per tool call. Merged rather than rebased, so nothing already reviewed has been rewritten.

`TestRun_ToolContextCarriesAttributes` guards the resolution, and was verified to fail when the re-applied line is removed.

## Follow-up, not in this PR

`adapter/a2a/executor.go` still passes a hardcoded `""`, so an agent driven through the A2A server asserts nothing yet. Closing that needs a decision about which field carries the human (an executor option taking a `func(context.Context) string` resolver is the shape I'd suggest), so it is deliberately left out here.

## Testing

- `agent`: attribute container semantics — read/write, snapshot-is-a-copy, empty rejected, separate from metadata, and a concurrency test since tool interceptors write to one invocation in parallel.
- `agent` context: transport round-trip, empty-inherits, and copy isolation in both directions.
- `llmagent`: the invocation's attributes reach the tool *and* tool interceptors via ctx — the root of the propagation chain.
- `agenttool`: a sub-agent inherits its caller's attributes, transitively through parent→mid→inner, asserts nothing when the caller did, and cannot write back into the caller's set. **Negative-checked**: with the `agenttool` change reverted, the inheritance and transitive tests both fail.
- `runner`: the userID argument reaches the invocation (it was accepted and discarded for a long time, so this is the guard that keeps it wired), option delivery, userID-argument-wins precedence in both directions, and empty values ignored.
- `plugins/otel`: `user.id` on all three span kinds, arbitrary caller attributes on all three, and no key at all when nothing was asserted.
- `task lint:new` (0 issues), `task license:check`, `go build ./...`, `go vet ./...`, `gofmt` all clean; `-race` clean on every affected package.
- Full `-short` suite run with and without these changes on this branch: identical failure sets, all invalid-API-key and Docker-dependent tests unrelated to this change.


agent/metadata.go
```diff
@@ -138,6 +138,13 @@ type InvocationMetadata struct {
 	turn         int
 	totalUsage   llm.TokenUsage
 
+	// userID is the caller-asserted end user this invocation runs for, as
+	// passed to Runner.Run. Opaque and unverified: an application asserts
+	// whatever it knows about its user - an email, an internal id - and
+	// nothing here validates it. Observability-only; never an authorization
+	// input.
+	userID string
```
weeco: Wondering if we should rather have an `attributes map[string]string` instead as we probably don't want to expand attributes infinitely here. We can make the key wellknown with an AttrUserID constant or so. If a user also wants to log additional attributes only they can know (e.g. premium user vs normal) they can just add these and we can add them to this list and put the tags alongside.

agent/metadata.go
```diff
@@ -187,6 +209,17 @@ func (m *InvocationMetadata) InvocationID() string {
...
+// UserID returns the caller-asserted end user this invocation runs for, or
+// "" when the caller asserted none.
+//
+// The value comes from Runner.Run's userID argument and is exported by the
+// otel plugin as the user.id span attribute. It is opaque - an email, an
+// internal id, whatever the calling application knows - and unverified: this
+// SDK never validates it and never reads it for authorization.
+func (m *InvocationMetadata) UserID() string {
+	return m.userID
+}
```
weeco: If we ride on top SetMetadata() - which I'd probably recomemnd, we should update this method accordingly. Also see the SetMetadata() method which already includes an example for setting extra attributes and was made for this purpose

plugins/otel/genai/semconv.go
```diff
@@ -61,6 +61,14 @@ const (
...
 	// true or leave unset, never false.
 	AttrGenAIConversationCompacted = "gen_ai.conversation.compacted"
+
+	// AttrUserID is the OTel registry's user.id attribute: the end user an
+	// invocation runs for, as asserted by the calling application. Not a
+	// gen_ai key - the GenAI conventions define no user attribute - but it
+	// is where the ecosystem converged: OpenInference stamps it, Langfuse's
+	// SDK writes it, ADK imports the same constant. The value is opaque (an
+	// email, an internal id) and verified by nobody.
+	AttrUserID = "user.id"
```
weeco: NIT: A bit odd to put the attr here in this pkg (genai conv whereas this is not genai), but not sure if there's a better or more natural place to put this

agent/metadata.go
alenkacz: I feel like user id is pretty critical field, I would say it's fine as top level. If you feel strongly, I can move to map

malinskibeniamin (approved):

alenkacz:
Added one commit since approval — `feat(a2a): resolve the request's end user for Runner.Run`. Flagging it because it re-opens review.

**Why it belongs here rather than in a follow-up.** Everything above gives the SDK the *ability* to carry an end user, but `adapter/a2a/executor.go` still passed a hardcoded `""`, so an agent served over A2A asserted nothing regardless. That is the managed-deployment path, and it is the one hop where the human genuinely *is* known — the server authenticates the request before the executor runs. Without this, the feature works only for agents that stamp `user.id` themselves and silently does nothing for A2A-served ones.

**Shape.** `WithUserIDResolver(func(context.Context) string)` — a callback, not a value, because the A2A protocol has no end-user field and one `Executor` serves every request while the user arrives per request. Injecting it also keeps this adapter independent of any particular auth package. `NewExecutor` gained variadic options, so all existing call sites compile unchanged, and the default (no resolver, or a resolver that finds nothing) asserts nothing exactly as before.

Consumer side looks like:

```go
a2a.WithUserIDResolver(func(ctx context.Context) string {
    id, ok := authentication.ExtractIdentity(ctx)
    if !ok {
        return ""
    }
    return id.EffectiveEmail()
})
```

Tests cover all four cases: no resolver, a resolver that resolves, a resolver that finds nothing, and a nil resolver. Full `golangci-lint` over `./adapter/a2a/...` is clean for the touched files (the remaining findings there are pre-existing in `processEvents` and `kvstore/taskstore_test.go`).

PR #225, opened 2026-09-15
feat(durable): durable execution for agents with Redpanda as the log

david-yu:
## Why

An agent run is a multi-step process that can span minutes or weeks: the model calls tools, tools call slow systems, and sometimes a human has to answer before the run can continue. Today, losing the process that started a run loses the work. Retrying from scratch re-charges the model, re-executes tools that already had side effects, and drops everything the run had established.

This adds durable execution whose only stateful dependency is Redpanda. No database, no coordination service, no second control plane. Every durable fact about a run is an append-only record keyed by run id, so the durability, replication, retention, and multi-consumer fan-out properties already relied on for event transport now apply to agent execution too.

Draft because the licensing decision below needs a call before this is mergeable.

## What

Three pieces, described in full in [`docs/durable-execution.md`](https://github.com/redpanda-data/ai-sdk-go/blob/dyu/durable-execution/docs/durable-execution.md).

- **`durable.Client`** starts and continues runs, delivers external input, and cancels, all by writing commands to a topic.
- **`durable.Worker`** hosts `agent.Agent` implementations, consumes tasks, and journals every message the agent appends and every tool result it produces. A replacement worker resumes from the last recorded step.
- **`durable/engine`** is the controller. It folds the journal into per-run state, dispatches tasks, enforces leases, retries with backoff, fires durable timers, matches external input to waiting tool calls, and publishes results for downstream consumers.

Writing an agent does not change. Register the interceptor and tools can suspend:

```go
ag, _ := llmagent.New("support", prompt, model,
    llmagent.WithTools(reg),
    llmagent.WithInterceptors(durable.NewInterceptor()),
)
```

```go
func (t refundTool) Execute(ctx context.Context, args json.RawMessage) (json.RawMessage, error) {
    if large(args) {
        return durable.WaitForInput("approval") // parks the run, frees the worker
    }
    return refund(ctx, args)
}
```

An operator answers whenever they get to it and the run picks up where it stopped:

```go
client.SendInput(ctx, "refund-1", "approval", map[string]any{"approved": true})
```

Input that arrives before the run reaches that step is buffered, not dropped, so there is no race between an approval and the run asking for it. `durable.Sleep` and `durable.SleepUntil` work the same way through a durable timer. Without the interceptor these helpers degrade to ordinary tool results and nothing else changes.

Also included:

- **`engine.Trigger`** starts one run per record on a business topic, keyed so replays cannot duplicate runs. Results are published back to a topic. That pair of topics is the entire contract between event streaming and agent execution; the engine never subscribes to business topics itself.
- **Weighted rollout records** pin new runs to a `(version, task queue)` while in-flight runs keep the version they started on. Shifting 5% to 100% is one compacted-topic write, and rollback is the same write with the old weights.
- **`store/session/journal`**, a `session.Store` backed by an append-only topic. It appends one small record per new message instead of rewriting the whole session, and `History` returns every record ever written, including messages that compaction later pruned from the working context.

## Topics

| Topic | Key | Cleanup | Holds |
|---|---|---|---|
| `<prefix>.commands` | run id | delete | every intent, from clients and workers |
| `<prefix>.journal` | run id | delete | the ordered record of what actually happened |
| `<prefix>.tasks.<queue>` | run id | delete | work handed to a worker pool |
| `<prefix>.state` | run id | compact | latest `RunState` per run |
| `<prefix>.results` | run id | delete | terminal results for downstream consumers |
| `<prefix>.config` | config key | compact | rollout records |

Commands and journal must have equal partition counts. The engine consumes commands in a consumer group and, on assignment of partition *p*, replays journal partition *p* to rebuild the runs living there. Sharding and failover come from consumer-group semantics rather than custom leader election.

## Change to an existing package

`detectIncompleteToolCalls` in `agent/llmagent` now also recognises a session whose tail is an assistant message with unanswered tool requests, and returns the index where the recovered tool-response message belongs. That is the shape a resumed run has, because it continues the same turn with no new user message. The runner's existing shape, an assistant tool request followed by a fresh user message, behaves exactly as before, and a table test covers both plus the negative cases.

## Licensing, needs a decision

Durable execution is set up as a Redpanda Enterprise feature: the new packages carry a header pointing at the Redpanda Community License Agreement in `licenses/RCL.md` rather than the Apache 2.0 license that covers the rest of the SDK. The boundary is confined to `durable/`, `durable/engine/`, `store/session/journal/`, and the example, and no Apache-licensed package in the repository imports them, so the split is mechanically enforceable. `licenses/README.md` documents it.

Two things worth a maintainer's call:

1. Whether a dual-licensed tree is acceptable in this repository at all, or whether enterprise code should live in a separate one.
2. `task license` adds the Apache header to any Go file missing one. The new files pass `license:check` because `addlicense` only looks for a copyright marker, but a stricter check later would need to know about the second header.

## Tests

Unit tests, no external dependencies:

- Engine state machine: idempotent starts, stale-token rejection, message ordering and duplicate sends, suspension with direct and buffered input, tool-result pruning, retry backoff through exhaustion, reopening a closed run, and a replay-equivalence test that folds a captured journal into a fresh state and compares.
- Worker: journals every message and tool result with the right indices, returns a journaled tool result without re-executing the tool, suspends on `WaitForInput` without calling the model again, resumes with delivered input, treats a journal write failure as retryable, and reports an interrupted shutdown as no outcome at all.

Integration test against a Redpanda container (`task test:durable`), all passing:

- a plain run completes and publishes a result
- human approval suspends the run, holds no worker, and resumes with the payload
- input sent before the wait is buffered and delivered
- **worker crash mid-tool** is recovered by lease expiry: a second worker re-executes only the unfinished tool, and journaled progress is not duplicated
- **engine restart** replays the journal: suspended runs are still suspended, closed runs survive, and the run then completes and accepts another turn

That crash test found a real bug worth calling out. When a worker is cancelled mid-tool, the agent fills the tool message with "interrupted, outcome unknown" placeholders. The worker was journaling that message, so the next attempt inherited a poisoned result. A shutting-down worker now journals nothing further: it reports no outcome, the lease expires, and the attempt is redone cleanly.

## Try it

```bash
docker compose up -d
cd examples/durable_agent
go run . engine &
go run . worker &
go run . start refund-1 "refund order 42"
go run . describe refund-1     # suspended, awaiting approval
go run . input refund-1 approval '{"approved":true,"by":"ops"}'
go run . wait refund-1
```

Kill the worker between `start` and `input`, then start it again. The run picks up where it left off.


david-yu:
Still a draft and an experiment for now.

PR #226, opened 2026-09-16
feat(catalog): Declare the publisher of every catalog offering (AI-2178)

MaciejWrzesniewski-Redpanda:
## Summary

Every offering in every ai-sdk-go catalog now declares the vendor that published it, under one shared attribute key. The console needs a model's brand from the catalog: [AI-2018](https://redpandadata.atlassian.net/browse/AI-2018) deletes `MODEL_BRAND_MATCHERS`, five regexes over the model ID, and has nothing to read in its place until these values exist. Vertex already declared a publisher; the other five catalogs did not.

The key is authored beside the model and never derived from an offering ID, and a test over every catalog the snapshot covers fails the build when an offering omits it. The repo's own add-model guide records the requirement for the next provider package.

## Review Load

| Bucket | Files # | Code Lines | Comment Lines |
|:---|:---:|:---:|:---:|
| Implementation | 10 | 71 | 82 |
| Tests | 5 | 147 | 105 |
| Docs / config | 3 | 535 | N/A |

`catalog/snapshot.json` is 512 of the 535 Docs / config lines - generator output, one `attributes` entry per offering, checked against a fresh run by `TestCommittedSnapshotIsFresh`. The comment split comes from classifying the added lines of the Go files, which is why the Markdown and JSON row carries none.

### Public Interface Delta

- `catalog`
  - `attributes.go`
    - [ADDED const `AttributePublisher` = `publisher`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/catalog/attributes.go#L44)
    - [ADDED consts `PublisherAmazon`, `PublisherAnthropic`, `PublisherGoogle`, `PublisherMeta`, `PublisherMistral`, `PublisherOpenAI`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/catalog/attributes.go#L52-L57) - the publisher values in use today, so a consumer branching on one compares against a symbol
    - [ADDED func `MustDeclarePublisher`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/catalog/attributes.go#L77) - panics with an `error` on an authoring error, and carries the `Must` prefix for it, matching `catalog.MustNew` and `catalog.MustDate`
- `providers/vertex`
  - `models.go`
    - [DEPRECATED const `ModelMetadataPublisher`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/providers/vertex/models.go#L94) - still the same value, now an alias for `catalog.AttributePublisher`

The key moved to the shared package because every catalog declares it - it is the one attribute that is not a provider quirk. `providers/vertex`'s `ModelMetadataPublisher` aliases it, same value, so callers that import it from there keep compiling; the `Deprecated` marker points them at the shared name.

The `Publisher*` consts ship beside the key so the vocabulary is a symbol rather than a doc-comment list. They do not close the set: a new vendor adds a const and its catalog declares it, the way `http.MethodGet` sits beside an arbitrary method string. Every publisher value in the repo now references one of them, so no vendor name is spelled as a literal outside `catalog`.

### The Bedrock publisher is authored per family, never split off an offering ID

Bedrock offering IDs carry an inference-profile geo prefix, so the vendor namespace belongs to the family's `BareID` and not to the ID a consumer sees. Authoring it on the family keeps it a declaration, and makes every variant of one family report the same vendor:

| Family `BareID` | Offering IDs it expands to | Publisher |
|---|---|---|
| `anthropic.claude-opus-5` | `global.`, `us.`, `eu.` and `au.` prefixed variants | `anthropic` |
| `amazon.nova-2-lite-v1:0` | `global.`, `us.`, `eu.` and `jp.` prefixed variants | `amazon` |

Neither family publishes a bare entry, so no offering ID in either row is its own `BareID` - a bare entry appears only where the family publishes one, as `mistral.mistral-large-3-675b-instruct` does. The ID split exists in one place only, and it is the test that re-derives the vendor to prove the authored value agrees with it.

### The vocabulary is open, not a closed set

[AI-2178](https://redpandadata.atlassian.net/browse/AI-2178) left this as an open decision. It ships open: a new vendor appearing in any catalog adds a value, and nothing asserts membership of a fixed list. A closed set would make a `switch` in the console look safe, which is the pattern [AI-2018](https://redpandadata.atlassian.net/browse/AI-2018) exists to remove. The six values in the committed snapshot:

| Publisher | Offerings | Catalogs it appears in |
|---|---|---|
| `anthropic` | 53 | `anthropic`, `aws.bedrock`, `gcp.vertex` |
| `openai` | 29 | `openai`, `aws.bedrock` |
| `google` | 15 | `google`, `aws.bedrock`, `gcp.vertex` |
| `amazon` | 4 | `aws.bedrock` |
| `mistral` | 1 | `aws.bedrock` |
| `meta` | 1 | `meta` |

103 offerings, 103 publishers. `mistral` has no console icon today - that is [AI-2018](https://redpandadata.atlassian.net/browse/AI-2018)'s to handle, and it is now visible in the catalog rather than hidden behind a regex that never matched it.

## Out of Scope

- The console reading `publisher` and deleting `MODEL_BRAND_MATCHERS` - [AI-2018](https://redpandadata.atlassian.net/browse/AI-2018).
- The Bedrock model picker's group headers, the console's second hand-kept vendor list - [AI-2179](https://redpandadata.atlassian.net/browse/AI-2179).
- The proto field and the aigw plumbing that carry the value out - [AI-2032](https://redpandadata.atlassian.net/browse/AI-2032), already merged as [cloudv2#29823](https://github.com/redpanda-data/cloudv2/pull/29823).
- Bumping cloudv2's `go.mod` to a revision containing this change, and moving its two `vertex.ModelMetadataPublisher` call sites onto `catalog.AttributePublisher` - [AI-2018](https://redpandadata.atlassian.net/browse/AI-2018) owns both, as bullets of its own Implementation Plan. The key cloudv2 reads is the plain string `publisher` and this change does not move it, so `ListModels` reports a publisher for all 99 offerings on the bump alone.
- The rollout order these values force: merge here, then cloudv2 bumps `go.mod` and aigw reaches production, then [AI-2018](https://redpandadata.atlassian.net/browse/AI-2018) removes `MODEL_BRAND_MATCHERS`. Shipping the console first renders a neutral glyph for all 103 offerings, which is the regression [AI-2018](https://redpandadata.atlassian.net/browse/AI-2018) exists to remove. Nothing but this line holds that order today.
- `openaicompat`, which has no catalog to declare against.
- Asserting the publisher values against an allowed set. The four single-vendor catalogs now have their values pinned to their provider name, so a misspelling fails - but nothing asserts membership of a fixed vocabulary, which is the section above's deliberate choice. No ticket, and none wanted while the vocabulary stays open.

### Acceptance Criteria

Coverage of [AI-2178](https://redpandadata.atlassian.net/browse/AI-2178), in the work item's own numbering.

- **AC 1** - all 99 offerings of `anthropic`, `aws.bedrock`, `google` and `openai` carry a non-empty `publisher` attribute.
  - [`TestEveryOfferingDeclaresAPublisher`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/cmd/catalog-snapshot/publisher_test.go#L35) - presence for every catalog, and the exact value for the four whose publisher is their provider name
  - [`TestCommittedSnapshotIsFresh`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/cmd/catalog-snapshot/snapshot_test.go#L33) - the committed snapshot carries the values a consumer will read: `anthropic` 12, `aws.bedrock` 50, `google` 11, `openai` 26
- **AC 2** - each Bedrock offering's publisher equals the vendor namespace of its family's `BareID`, so the bare, geo-prefixed and global variants of one family report the same publisher.
  - [`TestPublisherMatchesBareIDVendor`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/providers/bedrock/publisher_test.go#L41) - re-derives the vendor from each family's `BareID` and compares it with the authored `Publisher`, then checks that every expanded offering carries that publisher as a segment of its own ID
- **AC 3** - the publisher is authored on the catalog entry or its family declaration, and nothing in the catalog build path derives it by splitting an offering ID.
  - [`TestPublisherMatchesBareIDVendor`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/providers/bedrock/publisher_test.go#L41) - the only ID split in the change lives in this test, and production has to agree with it
  - [`TestMustDeclarePublisherGuards`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/catalog/attributes_test.go#L50) - an unauthored publisher fails at the authoring site rather than defaulting to something
- **AC 4** - a test fails when any offering in these catalogs has no publisher, so a newly added model cannot ship without one.
  - [`TestEveryOfferingDeclaresAPublisher`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/cmd/catalog-snapshot/publisher_test.go#L35) - it walks every catalog the snapshot covers, not the four this change touched
- **AC 5** - the Vertex catalog is unchanged: its three offerings still report `anthropic`, `anthropic` and `google`.
  - [`TestCommittedSnapshotIsFresh`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/cmd/catalog-snapshot/snapshot_test.go#L33) - the `gcp.vertex` block of the regenerated snapshot is identical to the one on `origin/main` at `d6d5e65e01`
- **AC 6** - a consumer reads the publisher by importing `catalog` alone: the key ships as an exported const in the shared `catalog` package, so naming it costs no import of a provider package.
  - [`catalog.AttributePublisher`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/catalog/attributes.go#L44) - the exported key, beside the six [`catalog.Publisher*`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/catalog/attributes.go#L52-L57) value consts; `vertex.ModelMetadataPublisher` stays as a deprecated alias so callers importing it from there keep compiling

### Additional Criteria

Authored for this change, not carried by [AI-2178](https://redpandadata.atlassian.net/browse/AI-2178).

- **A** - the `meta` catalog, which the work item does not name, also declares a publisher, and the guard covers whatever catalog is registered for the snapshot rather than a fixed list of four.
  - [`TestEveryOfferingDeclaresAPublisher`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/cmd/catalog-snapshot/publisher_test.go#L35)
- **B** - declaring a publisher across a single-vendor catalog cannot overwrite a different publisher already authored on an entry, and cannot declare an empty one; both are authoring errors and both fail immediately.
  - [`TestMustDeclarePublisherGuards`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/catalog/attributes_test.go#L50)
- **C** - attributes an entry already carries survive the declaration, so a provider's own discovery flags are not dropped by it.
  - [`TestMustDeclarePublisherDeclaresOnEveryEntry`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/catalog/attributes_test.go#L30)
- **D** - a misspelled publisher constant in a single-vendor catalog fails the build rather than shipping a value no console icon matches, and a catalog listed as neither single-vendor nor multi-vendor fails as well.
  - [`singleVendorProviders`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/cmd/catalog-snapshot/publisher_test.go#L69) - the publisher of `anthropic`, `google`, `meta` and `openai` must equal the provider name; `aws.bedrock` and `gcp.vertex` are listed as multi-vendor and have their own value checks, and a catalog in neither list fails the test rather than being skipped by it
- **E** - a Bedrock family authored without a `Publisher` fails at the authoring site, where no declaration in `bedrockFamilies` can reach the guard.
  - [`TestExpandFamiliesRequiresPublisher`](https://github.com/redpanda-data/ai-sdk-go/blob/c8bbcff33461ddb976786541d2f214ac9955282e/providers/bedrock/families_test.go#L30)

## Related Resources

- [AI-2178](https://redpandadata.atlassian.net/browse/AI-2178) - the work item this closes.
- [AI-2018](https://redpandadata.atlassian.net/browse/AI-2018) - the console change these values unblock.
- [AI-2179](https://redpandadata.atlassian.net/browse/AI-2179) - the Bedrock picker's group headers, the other hand-kept vendor list.
- [AI-2032](https://redpandadata.atlassian.net/browse/AI-2032) and [cloudv2#29823](https://github.com/redpanda-data/cloudv2/pull/29823) - the proto field and aigw plumbing that already carry `publisher` out.

[AI-2018]: https://redpandadata.atlassian.net/browse/AI-2018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AI-2178]: https://redpandadata.atlassian.net/browse/AI-2178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MaciejWrzesniewski-Redpanda:
@claude review

MIGRATION.md
```diff
@@ -101,6 +102,15 @@ shape.
   retired offerings remain (with `Life.Retires` in the past and a
   `Deprecated:` marker on their ID constants) so historical usage stays
   priceable. Use `Catalog().Now()` views to filter by lifecycle.
+- **`catalog.New` rejects an entry with no publisher.** Every offering
```
weeco: Wondering if we should delete the enitire migration.md. This seems mostly relevant for sdk internal development and not for users. Feels a bit odd to maintain this migration.md given the state of this SDK where we do breaking changes regularly. WDYT?

catalog/publisher.go
```diff
@@ -0,0 +1,53 @@
...
+	PublisherGoogle    = "google"
+	PublisherMeta      = "meta"
+	PublisherMistral   = "mistral"
+	PublisherOpenAI    = "openai"
+)
+
+// MustDeclarePublisher returns a copy of entries with [Entry.Publisher]
+// set to publisher. It panics on an empty publisher, or on an entry that
+// already declares a different one.
+func MustDeclarePublisher(publisher string, entries []Entry) []Entry {
```
weeco: A publisher should not be part of a `catalog.Entry`, but rather part of a `catalog.Fact`. A fact is a model card e.g. for Opus 4.8 , the entry is the inference provider (bedrock, anthrophic, openrouter, ...) specific catalog information. 

Given that the publisher string must not be empty (otherwise panic) we should define an alias type instead so this is not possible to do in the first place. Quite similar to the ModelID, ProviderID and ReasoningEffort

MIGRATION.md
MaciejWrzesniewski-Redpanda: I agree. As this is out of the scope of this PR and Jira item, should I file the Jira issue for this or should we discuss this with the team first?

catalog/publisher.go
MaciejWrzesniewski-Redpanda: Thanks for raising this. I was in a blind spot, because the version you reviewed was after a heavy refactoring and along the way I lost the big picture of the code.

The stuff is updated, please re-review.

But being honest with you here, this catalog.entry and catalog.fact nomenclature is not easily mapping in my head to LLM Model's Lab/Host (or Creator/Vendor?) split 😅

MIGRATION.md
weeco: Just delete the file as part of this PR, it's no problem

weeco (approved):

MIGRATION.md
MaciejWrzesniewski-Redpanda: Done in aa515e9. The two breaking changes it recorded are listed in the PR description now.

PR #227, opened 2026-09-18
DEVPROD-4812: enroll in AI-approved merge pilot (dry-run)

rpdevmp:
Enrolls ai-sdk-go in the DevProd AI-approved-merge pilot (DEVPROD-4812) in **dry-run**: every PR is evaluated and gets an audit comment, but nothing is approved or merged.

**Why the mechanism is vendored here:** this repo is public and GitHub does not allow a public repo to use reusable workflows or actions from a private repo (devprod-infra). So `.github/actions/ai-merge/` is a **vendored copy** of the reviewed mechanism (devprod-infra PR #816, source commit `6f3fd35` + org-membership fix from devprod-infra #829), with re-sync instructions in its README. devprod-infra remains the canonical source.

**Security model:** the workflow checks out the mechanism from the PR's **base ref** into `./mechanism` and runs it from there; the PR head is never checked out, so a PR cannot modify the guardrails that judge it. The guardrail config (`.github/ai-merge.yml`) is likewise read from `main`.

**Model (per ADP's request):** the repo is the opt-in; no per-PR label. Add `ai-merge-skip` to exclude a PR.

**Repo-specific exclusions** on top of the baseline (`.github/**`, `terraform/**`, release/container tooling, iam/auth dirs, credential/secret/key files): `providers/*/provider.go` (client construction / credentials), `Taskfile.yaml`, `taskfiles/**`, `.golangci.yaml` (they define what CI enforces), `LICENSE`, `header.txt`. Catalog adds and dependency bumps (with supply-chain checks) are in scope.

**CI on this PR:** the mechanism isn't on `main` yet, so the workflow detects that and skips with a notice (green). `test-ai-merge` runs the 26 mechanism tests. Go/lint/license checks are path-filtered and don't apply to a `.github/`-only change.

**Go-live** is a one-line change (`AI_MERGE_DRY_RUN: "false"`) once dry-run verdicts look right on real PRs. Requires one human approval.


rpdevmp:
@claude review once

rpdevmp:
## Review response — Copilot (5) and Claude (8 + notes) all addressed in the latest push

Thanks both — several of these were real go-live issues. Applied here in the vendored copy; upstreaming to the canonical mechanism in devprod-infra is a recorded follow-up (before any private repo enrolls). 38 tests.

**Trust boundary**
- **`pull_request` → `pull_request_target`** (Copilot #3 / Claude doc note). The workflow that holds the App key is now read from the base branch; a PR can no longer edit it to reach secrets or approve itself. Safe because the PR head is never checked out or executed. Bootstrap: this PR won't run the workflow (not on `main` yet); evaluation starts on the next PRs after merge.
- **Base must be the default branch** (Claude #1). `base.ref == repository.default_branch` in the job `if`, so nobody can push a branch with a modified mechanism and open a PR against it. Header comment states the invariant precisely, including the `base.sha`-at-last-sync caveat. CODEOWNERS on `.github/workflows/**` noted as a follow-up for ADP.
- **Drafts** (Claude #2): `draft == false` in the `if`; `ready_for_review` added to trigger types.

**Opt-out / approvals**
- **Skip label after approval** (Copilot #4 / Claude #3): no longer a job gate. Passed into the action as `skipped`, which abstains with an explicit reason, **withdraws the bot's prior approval and runs `gh pr merge --disable-auto`**. Label events re-run only for `ai-merge-skip`, so other labels don't trigger a paid model call.
- **Head-SHA race**: `::notice::` + clean exit instead of a red job; the new push re-evaluates.

**Reviewability / validation**
- **Binary or patchless files** (Copilot #1 / Claude #5): `files.json` carries `has_patch`; any file without a reviewable patch makes the PR ineligible (pure deletions and zero-change renames excepted). Tests added.
- **`confidence: true`** (Copilot #2): `_unit_float` accepts only `int`/`float`; bool and strings rejected. Regression test added.
- **Non-numeric config** (Claude #8): numeric config keys fail closed with a named reason instead of a crash; entries without `filename` likewise. Tests added.
- **`mantle.go`** (Claude #4): excluded explicitly (SigV4 transport retrieving/validating AWS credentials); `providers/*/provider.go` → `providers/**/provider.go` for nested packages. Config `version: 2`.
- **Fence** (Claude note): closer neutralisation is case-insensitive and whitespace-tolerant. Test added.

**Supply chain**
- **Pinned** `PyYAML==6.0.3` and all actions (`checkout`, `setup-python`, `create-github-app-token`) to commit SHAs with the tag in a comment (Claude #6).

**Audit trail**
- **Sticky comment** matched by the bot's own login **and** the marker, with `| head -1` for `--paginate` per-page output (Copilot #5 / Claude #7).
- **Anthropic call**: two retries with backoff on 429/5xx/network; `stop_reason == max_tokens` surfaced in the abstain reason.
- All `${{ }}` in `run:` blocks moved to `env:`.

**Tests** (Claude's two gaps): `test_repo_config.py` loads this repo's real `.github/ai-merge.yml` and asserts credential/CI-defining paths are excluded, catalog work stays eligible, and `go.mod` is a dependency change; `confidence_threshold` from config is now exercised. 38/38 in CI via `test-ai-merge`.

**Answered, not changed:** GitHub App bots are skipped at the gate, so "dependency bumps in scope" means human-authored or member bot *user* accounts (e.g. docs' `vbotbuildovich`) — stated explicitly in the README and config.

rpdevmp:
@claude review once

rpdevmp:
## Update: approval-only + risk-aware sizing (`305b53a`)

Two changes since the last review round, both driven by feedback in the Slack thread.

### 1. The bot never merges (per @weeco, confirmed)
- The approve step now posts a **binding approval pinned to the reviewed commit** and stops. There is no `gh pr merge` anywhere in the mechanism (the apply step asserts it).
- The withdraw step no longer touches auto-merge state — that's the author's own per-PR setting. It still dismisses the bot's prior approval when a later run doesn't re-approve (new commits, opt-out label, or a failed run), and runs `always()` so an infrastructure failure can't leave a stale approval standing.
- Merging is the author's call, or GitHub's native "Enable auto-merge", which respects the repo's required checks.
- Consequence: the App's `contents: write` is gone. Installed permissions are now **contents: read, members: read, metadata: read, pull_requests: write** — the minimum the mechanism can run with.

### 2. Size caps measure attention, not risk
- `max_changed_files` / `max_total_lines` now apply to **reviewable (non-generated) lines only**. A repo may declare `generated_paths`: machine-produced files that a **required** CI check regenerates and compares (so a hand edit fails CI and cannot merge). Those are exempt from the caps and stripped from the diff the model sees; the model is told they changed. An excluded path always wins over a generated path; `max_generated_lines` (20000) is a sanity ceiling; generated files must still be text.
- Every audit now records composition — reviewable / generated / test / source lines, tests-changed-with-source — as inputs for a separate risk-scoring layer. Informational in v1, not gating.
- **`generated_paths` is empty in this repo on purpose.** `catalog/snapshot.json` is the candidate, but `catalog:check` doesn't run on PRs today (`test.yaml` runs `task test`, not `task ci`, and its path filter skips `.json`-only changes), so exempting it would be unsafe. Two small ADP asks, both noted in the config: run `catalog:check` on PRs incl. `catalog/**`, and mark it required. Then we add `catalog/snapshot.json` and bump `version`.
- Also excluded `tool/mcp/transport.go` (spawns MCP servers via `exec.Command`), per the earlier Claude note.

### Verification
- 48/48 tests in CI (45 unit + 3 against this repo's real `.github/ai-merge.yml`).
- 23 realistic PR shapes simulated against the actual config: the #215-style catalog add and a clean dependency bump approve; `provider.go`, `mantle.go`, `transport.go`, `Taskfile.yaml`, `.github/**`, `terraform/**` refuse on excluded path; binary files refuse; non-members and a broken membership check (403) refuse with distinct reasons; a dep bump with no supply-chain object or a flagged install script abstains; bool / `NaN` / 0.6 confidence and `request_changes` abstain; 62 files and 1200 reviewable lines refuse; `min_confidence: 95` refuses instead of silently loosening. The 5000-line snapshot regen refuses **today** and approves once `generated_paths` is enabled — the intended behavior.
- Public-repo safety unchanged: fork PRs never run; same-repo authors are verified as org members via the API (`octocat` → 404, `rpdevmp` → 204 confirmed live); non-members couldn't merge regardless. A fork-PR probe will confirm the first point live after merge.

### Still the case
Dry-run (`AI_MERGE_DRY_RUN: "true"`): evaluates and comments, approves nothing. `pull_request_target` from the default branch only; PR head never checked out; membership via API; drafts excluded; opt-out handled in-run; pinned actions and `PyYAML==6.0.3`.

rpdevmp:
@claude review once

rpdevmp:
Round 3: all four taken, plus every note.

1. **`.claude/**`, `.claude-pr/**`, `CLAUDE.md` excluded.** Agreed — `settings.json` runs shell hooks on every contributor's machine; same category as Taskfile. Covered by the real-config test.
2. **Rename-into-generated.** Closed the asymmetry: a file is generated only if its current name *and* (when present) its previous name match `generated_paths`; `strip_generated` now matches the new path only. Tests: a hand-written file renamed into `docs/generated/**` stays reviewable (and trips the cap); a rename *within* generated paths stays generated.
3. **`.json`-only PRs run no CI.** You're right that the config comment implied `catalog:check` was the safeguard. It now states the actual situation plainly: the ruleset has **no required status checks**, all Go workflows path-filter on `**.go`/`go.mod`/`go.sum`, so a small hand edit to `snapshot.json` would have the AI review as its only gate today. Framed as a **go-live precondition for ADP** (run `catalog:check` on PRs incl. `catalog/**`; make Test / Lint / License / catalog:check required, ideally unconditional), not merely a `generated_paths` one.
4. **`tool/builtin/webfetch/**` excluded** (SSRF guard: deny list, rebinding-safe dialer, redirect hook). Covered by the real-config test.

Notes: **never-merges is now a test** (`test_mechanism_never_merges` reads `action.yml` and asserts no `gh pr merge` / `--auto` / `merge_method` / auto-merge mutation) so a re-sync can't quietly reintroduce it — and yes, my earlier "the apply step asserts it" referred to the deploy script, not the repo; the test fixes that properly. `_as_number` rejects non-finite floats with a named reason (`.inf`/`.nan` test). `strip_generated` runs once in `main()`; `build_user_prompt` documents the contract. `test_repo_config` skips visibly under pytest when no repo config exists. `catalog/provider.go` called out in the config comment as intentionally *not* matched. `glob_to_regex` is `lru_cache`d.

52/52 tests (49 unit + 3 real-config). Config `version: 4`.

.github/workflows/ai-approved-merge.yml
```diff
@@ -0,0 +1,99 @@
...
+  pull_request_target:
+    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
+
+permissions:
+  contents: read
+  pull-requests: read
+
+env:
+  AI_MERGE_DRY_RUN: "true"
+  AI_MERGE_MODEL: claude-sonnet-5
```
weeco, line 36: Sonnet models are expensive per task and not as smart as opus models

rpdevmp:
**Option B (agreed with @weeco in Slack): size is not a risk score, so it no longer gates.**

- Removed `max_changed_files` / `max_total_lines` / `max_generated_lines`. File and line counts are still recorded in every audit comment as signals for the separate risk-scoring work; they don't decide anything.
- The only size rule left is that the whole reviewable diff must **fit in one review**: `max_diff_chars` (per-repo, default 120000 ≈ 3000 lines). Larger changes abstain with *"too large to review in one pass"* and get a normal human review.
- The model must now return `reviewed_fully: true` — an approval is never given without it — and a `scope` classification (`mechanical` / `localized` / `broad`), recorded in the audit.
- `generated_paths` stays (spends the model's budget on hand-written code); still empty here pending `catalog:check` on PRs.
- Model switched to `claude-opus-5`; all internal references (private repo, ticket, PR numbers) removed from the public files.

Verified 16 realistic shapes against this repo's config: catalog add and clean dep bump approve; a 1200-line hand-written change and a 62-file change now approve (they fit in one review) — the intended consequence of "size ≠ risk"; a 130K-char diff abstains on the one remaining rule; every excluded-path / non-member / binary / opt-out / malformed-verdict / config-typo case still refuses. 52/52 tests. Config `version: 6`.

Follow-ups (own tickets): risk-scoring POC + design, fed by the composition/scope signals this dry-run will collect; phase-2 agentic reviewer as a separate unprivileged job handing its verdict to the approving job.

rpdevmp:
@claude review once

